Issue #86, opened 2026-05-27
Error 3-1: implicit cast not available between Number and Item in equality comparisons


ruizmaa:
`binary_implicit_type_promotion` raised 3-1 when comparing a `Number` cell against an `Item` constant (e.g. `{c0070} = [eba_CT:x212]`), even though comparison operators have a fixed `Boolean` return type and the mismatch is handled gracefully at runtime.

## Examples

```
with{tB_05.01, default:0, interval: false}: if ({c0070} = [eba_CT:x212]) then ({c0040} in {[eba_qCO:qx2000], ...})
with{tB_05.01, default:null, interval: false}: if ({c0070} = [eba_CT:x212]) then ({c0040} in {[eba_qCO:qx2000], ...})
```

Issue #84, opened 2026-05-27
Error 3-4: interval flag rejected for non-numeric cells


ruizmaa:
`with {col, interval: true}` propagates `interval: true` to all inner VarIDs,
including non-numeric ones (Item, Boolean, TimeInterval). `from_database_to_scalar_types`
raised 3-4 for any non-Number type receiving `interval: true`, even though the flag
is irrelevant for those types and should be silently ignored.

## Examples

```
with {c0010, interval: true}: if ( {tF_00.01, r0010, default: ""} = [eba_AS:x2] ) then ...
with {c0010, interval: true}: if ( {tF_00.01, r0010, default: ""} = [eba_qAS:qx2004] ) then ...
```

Issue #83, opened 2026-05-27
Error 3-6: Integer default rejected for non-Boolean cells


ruizmaa:
`__check_default_value` only accepted a default when a direct promotion path existed
between default type and cell type. `Integer 0` used as a null sentinel for `Item` or
`TimeInterval` cells has no such path, causing spurious 3-6 errors.

## Examples

```
with {tB_01.02, default: 0, interval: false}: if (not (isnull ({c0110}))) then ...
with {tB_02.02, default: 0, interval: false}: {c0080} > {c0070}
```

Issue #80, opened 2026-05-27
Error UNKNOWN: unhandled crashes in the semantic analyser


ruizmaa:
Two unhandled exceptions raised as UNKNOWN errors instead of producing a valid result.

## Examples

```
{tB_05.01, c0070} in { [eba_CT:x212], [eba_CT:x213] }
{tB_05.01, c0080} in { [eba_GA:AL], [eba_GA:AT], ... }
with {tR_04.00.a, c*, default: 0, interval: true} [where qPYB = [eba_qIA:qx2090]]: {r0100} + {r0060} >= 9000000 * {r0040}
```

PR #94, opened 2026-05-28
Semantic errors

ruizmaa:
## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/operators/base.py
+++ b/src/dpmcore/dpm_xl/operators/base.py
@@ -350,7 +350,8 @@ class Binary(Operator):
             )
             if left.records is not None and right.records is not None:
                 if (
-                    len(left.records) != len(right.records)
+                    not cls.do_not_check_with_return_type
+                    and len(left.records) != len(right.records)
                     and len(left.records) != 0
                     and len(right.records) != 0
                 ):
--- a/src/dpmcore/dpm_xl/operators/base.py
+++ b/src/dpmcore/dpm_xl/operators/base.py
@@ -297,9 +297,12 @@ class Binary(Operator):
         interval_allowed = getattr(cls, "interval_allowed", False)
         if isinstance(left_type, Mixed) or isinstance(right_type, Mixed):
             if result_dataframe is None:
-                raise Exception(
-                    "Mixed type promotion requires a result dataframe"
-                )
+                # Semantic-only mode: no row data available for per-row type
+                # resolution.  Return the operator's fixed return type when
+                # known; otherwise Mixed (the column stays unresolved).
+                return (
+                    return_type if return_type is not None else Mixed()
+                ), None
             final_type, result_dataframe = (
                 binary_implicit_type_promotion_with_mixed_types(
                     result_dataframe=result_dataframe,
--- a/src/dpmcore/dpm_xl/semantic_analyzer.py
+++ b/src/dpmcore/dpm_xl/semantic_analyzer.py
@@ -202,18 +202,20 @@ class InputAnalyzer(ASTTemplate, ABC):
         if default_value is None:
             return
         default_type = ScalarFactory().scalar_factory(code=default_value.type)
-        # Import the implicit type promotion dictionary for unidirectional check
         from dpmcore.dpm_xl.types.promotion import implicit_type_promotion_dict
 
-        # Check if default_type can be promoted TO type_ (unidirectional check)
-        # Get the set of types that default_type can be promoted to
-        default_implicities = implicit_type_promotion_dict[
-            default_type.__class__
-        ]
-
-        # If expected type is not in the set of types default can be promoted to,
-        # raise a semantic error
-        if not type_.is_included(default_implicities):
+        default_implicities = implicit_type_promotion_dict.get(
+            default_type.__class__, set()
+        )
+        cell_implicities = implicit_type_promotion_dict.get(
+            type_.__class__, set()
+        )
+        # Accept bidirectionally: D->C or C->D. Mixed has no dict entry; empty
+        # cell_implicities short-circuits the check (any default is valid).
+        if cell_implicities and not (
+            type_.is_included(default_implicities)
+            or default_type.is_included(cell_implicities)
+        ):
             raise errors.SemanticError(
                 "3-6", expected_type=type_, default_type=default_type
             )
--- a/src/dpmcore/dpm_xl/semantic_analyzer.py
+++ b/src/dpmcore/dpm_xl/semantic_analyzer.py
@@ -47,6 +47,7 @@ from dpmcore.dpm_xl.symbols import (
     Structure,
 )
 from dpmcore.dpm_xl.types.scalar import (
+    Boolean,
     Integer,
     Item,
     Mixed,
@@ -210,11 +211,16 @@ class InputAnalyzer(ASTTemplate, ABC):
         cell_implicities = implicit_type_promotion_dict.get(
             type_.__class__, set()
         )
-        # Accept bidirectionally: D->C or C->D. Mixed has no dict entry; empty
-        # cell_implicities short-circuits the check (any default is valid).
+        # Accept: D→C, C→D, or shared promotion target (excl. Boolean crossings).
+        # Mixed has no dict entry; empty cell_implicities skips the check.
         if cell_implicities and not (
             type_.is_included(default_implicities)
             or default_type.is_included(cell_implicities)
+            or (
+                not isinstance(default_type, Boolean)
+                and not isinstance(type_, Boolean)
+                and default_implicities & cell_implicities
+            )
         ):
             raise errors.SemanticError(
                 "3-6", expected_type=type_, default_type=default_type
--- a/src/dpmcore/dpm_xl/types/promotion.py
+++ b/src/dpmcore/dpm_xl/types/promotion.py
@@ -221,30 +221,45 @@ def binary_implicit_type_promotion_with_mixed_types(
         )
 
     elif isinstance(left_type, Mixed):
-        # Series.apply stubs don't know ScalarType is a valid cell value
-        result_dataframe["data_type"] = result_dataframe["data_type"].apply(
-            lambda x: binary_implicit_type_promotion(  # type: ignore[arg-type,return-value]
-                x,
-                right_type,
-                op_type_to_check,
-                return_type,
-                interval_allowed,
-                error_info,
+        if return_type is not None:
+            # Operator has a fixed return type (e.g. Boolean for comparisons).
+            # Per-row promotion would raise 3-1 for rows whose actual type
+            # cannot be promoted to the RHS type (e.g. Number vs Item), but
+            # at runtime such comparisons simply return false/null rather than
+            # an error.  Assign the known return type directly.
+            result_dataframe["data_type"] = return_type  # type: ignore[call-overload]
+        else:
+            # Series.apply stubs don't know ScalarType is a valid cell value
+            result_dataframe["data_type"] = result_dataframe[
+                "data_type"
+            ].apply(
+                lambda x: binary_implicit_type_promotion(  # type: ignore[arg-type,return-value]
+                    x,
+                    right_type,
+                    op_type_to_check,
+                    return_type,
+                    interval_allowed,
+                    error_info,
+                )
             )
-        )
 
     elif isinstance(right_type, Mixed):
-        # Series.apply stubs don't know ScalarType is a valid cell value
-        result_dataframe["data_type"] = result_dataframe["data_type"].apply(
-            lambda x: binary_implicit_type_promotion(  # type: ignore[arg-type,return-value]
-                left_type,
-                x,
-                op_type_to_check,
-                return_type,
-                interval_allowed,
-                error_info,
+        if return_type is not None:
+            result_dataframe["data_type"] = return_type  # type: ignore[call-overload]
+        else:
+            # Series.apply stubs don't know ScalarType is a valid cell value
+            result_dataframe["data_type"] = result_dataframe[
+                "data_type"
+            ].apply(
+                lambda x: binary_implicit_type_promotion(  # type: ignore[arg-type,return-value]
+                    left_type,
+                    x,
+                    op_type_to_check,
+                    return_type,
+                    interval_allowed,
+                    error_info,
+                )
             )
-        )
 
     if return_type:
         return return_type, result_dataframe
--- a/src/dpmcore/dpm_xl/types/promotion.py
+++ b/src/dpmcore/dpm_xl/types/promotion.py
@@ -43,6 +43,7 @@ implicit_type_promotion_dict: dict[type[ScalarType], set[type[ScalarType]]] = {
         Item,
         Subcategory,
         Null,
+        Mixed,
     },
 }
 
--- a/src/dpmcore/dpm_xl/types/promotion.py
+++ b/src/dpmcore/dpm_xl/types/promotion.py
@@ -161,10 +161,7 @@ def binary_implicit_type_promotion(
                 add_semantic_warning(
                     f"Implicit promotion between {left} and {right}."
                 )
-        if return_type and (
-            left.is_included(right_implicities)
-            or right.is_included(left_implicities)
-        ):
+        if return_type:
             return return_type
         elif left.is_included(right_implicities):
             return left
--- a/src/dpmcore/dpm_xl/ast/constructor.py
+++ b/src/dpmcore/dpm_xl/ast/constructor.py
@@ -119,7 +119,11 @@ class ASTVisitor(dpm_xlParserVisitor):
     ) -> WithExpression:
         ctx_list = list(ctx.getChildren())
         partial_selection: VarID = self._visit(ctx_list[1])
-        expression: AST = self._visit(ctx_list[3])
+        # Body expression is always the last child.  When the optional
+        # [WHERE expression] block is present the token count grows by 4,
+        # so ctx_list[3] would land on the WHERE terminal rather than the
+        # body.  ctx_list[-1] is correct in both cases.
+        expression: AST = self._visit(ctx_list[-1])
         return WithExpression(
             partial_selection=partial_selection, expression=expression
         )
--- a/src/dpmcore/dpm_xl/ast/operands.py
+++ b/src/dpmcore/dpm_xl/ast/operands.py
@@ -215,6 +215,11 @@ class OperandsChecking(ASTTemplate, ABC):
             and getattr(self.partial_selection, header) is not None
         ):
             return
+        # If ALL operands omit the header, the expression applies to every
+        # instance of that dimension ("apply to all" semantics) — valid.
+        # Only raise 1-20 when SOME operands specify the header and SOME don't.
+        if all(getattr(node, header) is None for node in self.operands[table]):
+            return
         for node in self.operands[table]:
             if getattr(node, header) is None:
                 if header == "cols":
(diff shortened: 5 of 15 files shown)


src/dpmcore/dpm_xl/types/promotion.py
```diff
@@ -160,10 +161,7 @@ def binary_implicit_type_promotion(
                 add_semantic_warning(
                     f"Implicit promotion between {left} and {right}."
                 )
-        if return_type and (
-            left.is_included(right_implicities)
-            or right.is_included(left_implicities)
-        ):
+        if return_type:
```
andres-sole, line 164: **[#3 / CONFIRMED / high]** Changed from `if return_type and (compatible): return return_type` to unconditional `if return_type: return return_type`. Combined with `do_not_check_with_return_type=True` on every comparison operator and Match, **any** pair of types now silently returns Boolean — not just Number↔Item.

**Trigger:** `{tT, c_boolCell} = {tT, c_dateCell}` (Boolean vs TimeInterval), `{tItemCell} > {tDateCell}` (Item vs TimePeriod) — previously raised 3-1, now Boolean with no diagnostic; runtime NaN/false rows.

If the intent was to allow Number↔Item only, narrow the bypass to that pair or whitelist via a `cross_type_comparison_allowed` set.

src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -202,18 +203,25 @@ def __check_default_value(default_value: Any, type_: ScalarType) -> None:
...
+        )
+        # Accept: D→C, C→D, or shared promotion target (excl. Boolean crossings).
+        # Mixed has no dict entry; empty cell_implicities skips the check.
+        if cell_implicities and not (
+            type_.is_included(default_implicities)
+            or default_type.is_included(cell_implicities)
+            or (
+                not isinstance(default_type, Boolean)
+                and not isinstance(type_, Boolean)
+                and default_implicities & cell_implicities
```
andres-sole, line 222: **[#4 / CONFIRMED / high]** The shared-promotion-target admit clause (`default_implicities & cell_implicities`, excluding Boolean) passes for nearly every non-Boolean type pair because almost every ScalarType promotes to String. New tests explicitly add Integer default-for-Item / Integer default-for-TimeInterval as valid.

**Trigger:** `with {tT, c_itemCell, default: 0}: {c_itemCell} = [eba_X:Y]` — `Integer→{String,Number,Integer}`, `Item→{String,Item}`, intersection `{String}`, neither Boolean. No 3-6 raised. Combined with finding #3, missing rows are filled with literal `0` and compared to an Item signature, silently returning false.

Consider requiring the *direct* promotion target (not transitive through String).

src/dpmcore/dpm_xl/types/promotion.py
```diff
@@ -220,30 +218,45 @@ def binary_implicit_type_promotion_with_mixed_types(
...
-                x,
-                right_type,
-                op_type_to_check,
-                return_type,
-                interval_allowed,
-                error_info,
+        if return_type is not None:
+            # Operator has a fixed return type (e.g. Boolean for comparisons).
+            # Per-row promotion would raise 3-1 for rows whose actual type
+            # cannot be promoted to the RHS type (e.g. Number vs Item), but
```
andres-sole, line 224: **[#5 / CONFIRMED / high]** When left/right is Mixed and `return_type` is provided, this branch broadcasts `result_dataframe['data_type'] = return_type` and skips the per-row `binary_implicit_type_promotion .apply`. That .apply previously emitted implicit-promotion warnings AND raised 3-1 for incompatible row pairs.

**Trigger:** Mixed column with rows {Number, Item, Boolean} compared to scalar Number — previously per-row apply warned for Item/Boolean rows and raised 3-1 for fully-incompatible ones. Now every row's data_type is force-assigned Boolean; runtime NaN/false with no warning.

At minimum, run the per-row .apply for warning purposes and still assign return_type at the end.

src/dpmcore/dpm_xl/operators/base.py
```diff
@@ -297,9 +297,12 @@ def validate_types(
         interval_allowed = getattr(cls, "interval_allowed", False)
         if isinstance(left_type, Mixed) or isinstance(right_type, Mixed):
             if result_dataframe is None:
-                raise Exception(
-                    "Mixed type promotion requires a result dataframe"
-                )
+                # Semantic-only mode: no row data available for per-row type
+                # resolution.  Return the operator's fixed return type when
+                # known; otherwise Mixed (the column stays unresolved).
+                return (
```
andres-sole, line 303: **[#7 / CONFIRMED / medium]** `raise Exception('Mixed type promotion requires a result dataframe')` replaced with `return (return_type if return_type is not None else Mixed(), None)`. For arithmetic operators (`return_type=None`), this silently returns `(Mixed(), None)` instead of erroring — bypassing `op_type_to_check=Number` validation entirely.

**Trigger:** `BinPlus.validate_types(left=Scalar(Mixed), right=Scalar(Item), result_dataframe=None)` — Scalar+Scalar path produces (None, None) from `validate_structures`; previously hard-raised, now `(Mixed(), None)` silently.

If the goal was to avoid the hard `Exception`, raise `SemanticError` instead so callers can route it properly.

andres-sole (commented):
**[#1 / CONFIRMED / highest severity] `src/dpmcore/dpm_xl/ast/constructor.py:557` (`visitOpRef`) and the matching new `cellAddress: ... opRef ...` grammar branch.** The new `{oOp1, r010}` cellAddress feature is broken end-to-end: `visitOpRef` builds `VarID(operation='Op1', table=None, rows=['010'])`, but every `visit_VarID` still has `if not node.table: raise SemanticError('1-4')` (see `ast/operands.py:565`, `semantic_analyzer.py:visit_VarID`, `ml_generation.py:visit_VarID`). None were updated to consume the new `operation` field. **Trigger:** `SemanticService.validate('{oOp1, r010} = 5')` → 1-4 raised. `tests/unit/ast/test_operation_cell_address.py` only covers `SyntaxService().parse()`.

**[#2 / CONFIRMED / high] `src/dpmcore/dpm_xl/operators/clause.py:Sub.validate`.** `Sub.validate` mutates `operand.structure.components` in place. The chained-loop in `semantic_analyzer.visit_SubOp` threads the same `Structure` object through each iteration, and the returned RecordSet shares it with the original — including any cached `calculations_outputs` entry. **Trigger:** `calc := {tT}; calc[sub a=1, b=2]` or any two references to a calc with multi-sub. Deep-copy the structure before each `SubOperator.validate` call, or fix `Sub.validate` to return a fresh `Structure`.

**[#6 / CONFIRMED / medium] `src/dpmcore/dpm_xl/types/scalar.py:321` removed `interval` guard.** Deletion of `if interval: raise SemanticError('3-4', ...)` means non-numeric types now silently drop the `interval` flag. Only Number/Integer store it; for String/Boolean/Item/TimeInterval/Date/TimePeriod it disappears at the factory. **Trigger:** `with {tT, default: 0, interval: true}: {c0010} <= 1` where `tT.c0010` is String — previously raised 3-4, now `interval: true` becomes a no-op; the user's typo is hidden.

**[#7 / CONFIRMED / medium] `src/dpmcore/dpm_xl/operators/base.py:303` Mixed + None fallback.** *(also anchored inline below — duplicate listed here for the full chain)*. Original location: hard `raise Exception` replaced with `return (return_type if return_type is not None else Mixed(), None)`; for arithmetic operators with `return_type=None`, this silently returns `(Mixed(), None)`.

**[#8 / CONFIRMED / medium] `src/dpmcore/dpm_xl/model_queries.py:1292` range boundary check removed.** Removal of `_check_ranges_values_are_present` calls in `get_table_data`. In principle fine — but the single-range filter path in `data_handlers` (`len==1 and '-' in cell_elements[0]`) only emits SQL `BETWEEN` and never verifies that the endpoints exist as header codes. The multi-value `cells_not_found` check lives only in the else branch. **Trigger:** `{tT, r0010-rXXX, c0020} >= 0` where `rXXX` is not a real row — `BETWEEN '0010' AND 'XXX'` returns whatever rows lexicographically fall in that span; validation silently runs on the wrong slice instead of raising 1-2. Either restore the endpoint-existence check (DB-side this time, not post-query DataFrame), or document that ranges no longer enforce endpoint validity.

**[#9 / PLAUSIBLE / medium] `src/dpmcore/dpm_xl/ast/constructor.py:visitVarRef`.** Unconditional classification of any `v_*` token as `PreconditionItem` makes the new VAR_REFERENCE grammar alternative `VAR_REF_PREFIX '_' TABLE_CODE` unreachable as a regular VarRef. **Trigger:** variable `_FooBar` (legal per the grammar's new alternative) → `v_FooBar` parses as PreconditionItem instead of VarRef. Distinguish by context (precondition vs select position) or by a different token.

**[#10 / CONFIRMED / low] `src/dpmcore/dpm_xl/semantic_analyzer.py:visit_TimeShiftOp`.** Hardcodes `shift_number=0` to `TIME_SHIFT.validate` whenever the AST node isn't a literal `Constant`. The 4-7-4 type guard is correct, but the value passed is wrong for `time_shift({tT}, Q, -5)` (UnaryOp wrapper) and `time_shift({tT}, Q, 5*12)`. **Trigger:** `RecordSet.origin` becomes `time_shift(..., 0, ...)` — error messages and any consumer keying on origin conflate distinct shifts into '0'.

**[#11 / PLAUSIBLE / low] `src/dpmcore/services/ast_generator.py:_to_ref_period`.** Assumes the `t-` branch always carries a positive numeric suffix. A programmatically-constructed `Constant(value=-5)` produces internal `t-Q-5` → `T--5Q` (double-minus). Parser-built ASTs never trigger; unit fixtures / API users can.

**[#12 / PLAUSIBLE / low] `scripts/check_semantic.py:226` `.strip()` on non-str.** SQLite's `trim()` coerces non-text affinities, so the WHERE filter does not exclude rows where Expression is stored as INTEGER/BLOB. **Trigger:** legacy migration leaves a row with bytes/int affinity → `.strip()` raises AttributeError mid-pass.

**[#13 / cleanup] `src/dpmcore/dpm_xl/ast/constructor.py:57` `_strip_backticks`.** Defined but never used. Three sites inline `if text.startswith('`'): return text[1:-1]` (visitKeyNames, visitPropertyCode, visitTemporaryIdentifier). The inline form lacks the helper's `len >= 2` guard. Pick one form.

**[#14 / cleanup] shift_number → period-string conversion duplicated** across `src/dpmcore/services/ast_generator.py:_extract_time_shifts` and `src/dpmcore/dpm_xl/ast/module_dependencies.py:visit_TimeShiftOp`; the two have already drifted (one uses `sn.value` directly, the other wraps in `str()`). Extract a single helper.

**[#15 / cleanup] `scripts/check_semantic.py:222`.** Loop prints a PASS line for every row; for ~50k validations this dominates runtime and floods CI logs. Add `--quiet` or a progress-bar mode.

---

Full write-up is also in `PR_REVIEW_semantic-errors.md` at the repo root.

src/dpmcore/dpm_xl/types/promotion.py
ruizmaa: The fix was motivated by Number = Item comparisons only (issue #86, PR #87), as the tests show. Making the bypass unconditional is broader than intended, I will narrow it to cases where return_type is provided and the pair involves Number/Item specifically.

src/dpmcore/dpm_xl/semantic_analyzer.py
ruizmaa: Intentional. Real validations use default: 0 on Item/TimeInterval cells as a standardnull sentinel for missing rows (check PR #82). The current aproach is broader, but this way we avoid the errors from issue #83

src/dpmcore/dpm_xl/types/promotion.py
ruizmaa: Intentional. For comparison operators, a Mixed column with incompatible rows (e.g. Item compared to Number) produces false/null at runtime, not an error. The per-row .apply was raising 3-1 for those rows incorrectly (PR #78, issue #80). Anyways, the warning suppression point is right, ill fix it

src/dpmcore/dpm_xl/operators/base.py
ruizmaa: This is intencional, before this change, a Mixed operand raised a `Exception,` which ended as `UNKNOWN` in the validation report (issue #80, PR #78). Instead of that, returning `Mixed()` keeps the column unresolved rather than crashing and the type is resolved correctly when row data is available. On the `SemanticError` suggestion i agree, I will replace the `Exception` path with a proper `SemanticError`.

ruizmaa:
> **[#1 / CONFIRMED / highest severity] `src/dpmcore/dpm_xl/ast/constructor.py:557` (`visitOpRef`) and the matching new `cellAddress: ... opRef ...` grammar branch.** The new `{oOp1, r010}` cellAddress feature is broken end-to-end: `visitOpRef` builds `VarID(operation='Op1', table=None, rows=['010'])`, but every `visit_VarID` still has `if not node.table: raise SemanticError('1-4')` (see `ast/operands.py:565`, `semantic_analyzer.py:visit_VarID`, `ml_generation.py:visit_VarID`). None were updated to consume the new `operation` field. **Trigger:** `SemanticService.validate('{oOp1, r010} = 5')` → 1-4 raised. `tests/unit/ast/test_operation_cell_address.py` only covers `SyntaxService().parse()`.
> 
> **[#2 / CONFIRMED / high] `src/dpmcore/dpm_xl/operators/clause.py:Sub.validate`.** `Sub.validate` mutates `operand.structure.components` in place. The chained-loop in `semantic_analyzer.visit_SubOp` threads the same `Structure` object through each iteration, and the returned RecordSet shares it with the original — including any cached `calculations_outputs` entry. **Trigger:** `calc := {tT}; calc[sub a=1, b=2]` or any two references to a calc with multi-sub. Deep-copy the structure before each `SubOperator.validate` call, or fix `Sub.validate` to return a fresh `Structure`.
> 
> **[#6 / CONFIRMED / medium] `src/dpmcore/dpm_xl/types/scalar.py:321` removed `interval` guard.** Deletion of `if interval: raise SemanticError('3-4', ...)` means non-numeric types now silently drop the `interval` flag. Only Number/Integer store it; for String/Boolean/Item/TimeInterval/Date/TimePeriod it disappears at the factory. **Trigger:** `with {tT, default: 0, interval: true}: {c0010} <= 1` where `tT.c0010` is String — previously raised 3-4, now `interval: true` becomes a no-op; the user's typo is hidden.
> 
> **[#7 / CONFIRMED / medium] `src/dpmcore/dpm_xl/operators/base.py:303` Mixed + None fallback.** _(also anchored inline below — duplicate listed here for the full chain)_. Original location: hard `raise Exception` replaced with `return (return_type if return_type is not None else Mixed(), None)`; for arithmetic operators with `return_type=None`, this silently returns `(Mixed(), None)`.
> 
> **[#8 / CONFIRMED / medium] `src/dpmcore/dpm_xl/model_queries.py:1292` range boundary check removed.** Removal of `_check_ranges_values_are_present` calls in `get_table_data`. In principle fine — but the single-range filter path in `data_handlers` (`len==1 and '-' in cell_elements[0]`) only emits SQL `BETWEEN` and never verifies that the endpoints exist as header codes. The multi-value `cells_not_found` check lives only in the else branch. **Trigger:** `{tT, r0010-rXXX, c0020} >= 0` where `rXXX` is not a real row — `BETWEEN '0010' AND 'XXX'` returns whatever rows lexicographically fall in that span; validation silently runs on the wrong slice instead of raising 1-2. Either restore the endpoint-existence check (DB-side this time, not post-query DataFrame), or document that ranges no longer enforce endpoint validity.
> 
> **[#9 / PLAUSIBLE / medium] `src/dpmcore/dpm_xl/ast/constructor.py:visitVarRef`.** Unconditional classification of any `v_*` token as `PreconditionItem` makes the new VAR_REFERENCE grammar alternative `VAR_REF_PREFIX '_' TABLE_CODE` unreachable as a regular VarRef. **Trigger:** variable `_FooBar` (legal per the grammar's new alternative) → `v_FooBar` parses as PreconditionItem instead of VarRef. Distinguish by context (precondition vs select position) or by a different token.
> 
> **[#10 / CONFIRMED / low] `src/dpmcore/dpm_xl/semantic_analyzer.py:visit_TimeShiftOp`.** Hardcodes `shift_number=0` to `TIME_SHIFT.validate` whenever the AST node isn't a literal `Constant`. The 4-7-4 type guard is correct, but the value passed is wrong for `time_shift({tT}, Q, -5)` (UnaryOp wrapper) and `time_shift({tT}, Q, 5*12)`. **Trigger:** `RecordSet.origin` becomes `time_shift(..., 0, ...)` — error messages and any consumer keying on origin conflate distinct shifts into '0'.
> 
> **[#11 / PLAUSIBLE / low] `src/dpmcore/services/ast_generator.py:_to_ref_period`.** Assumes the `t-` branch always carries a positive numeric suffix. A programmatically-constructed `Constant(value=-5)` produces internal `t-Q-5` → `T--5Q` (double-minus). Parser-built ASTs never trigger; unit fixtures / API users can.
> 
> **[#12 / PLAUSIBLE / low] `scripts/check_semantic.py:226` `.strip()` on non-str.** SQLite's `trim()` coerces non-text affinities, so the WHERE filter does not exclude rows where Expression is stored as INTEGER/BLOB. **Trigger:** legacy migration leaves a row with bytes/int affinity → `.strip()` raises AttributeError mid-pass.
> 
> **[#13 / cleanup] `src/dpmcore/dpm_xl/ast/constructor.py:57` `_strip_backticks`.** Defined but never used. Three sites inline `if text.startswith('`'): return text[1:-1]`(visitKeyNames, visitPropertyCode, visitTemporaryIdentifier). The inline form lacks the helper's`len >= 2` guard. Pick one form.
> 
> **[#14 / cleanup] shift_number → period-string conversion duplicated** across `src/dpmcore/services/ast_generator.py:_extract_time_shifts` and `src/dpmcore/dpm_xl/ast/module_dependencies.py:visit_TimeShiftOp`; the two have already drifted (one uses `sn.value` directly, the other wraps in `str()`). Extract a single helper.
> 
> **[#15 / cleanup] `scripts/check_semantic.py:222`.** Loop prints a PASS line for every row; for ~50k validations this dominates runtime and floods CI logs. Add `--quiet` or a progress-bar mode.
> 
> Full write-up is also in `PR_REVIEW_semantic-errors.md` at the repo root.


1. Wasnt changed in this PR, already existed
2. Wasnt changed in this PR, already existed
6. This is intentional. `OperandsChecking.visit_VarID` copies `interval: true` into every inner `VarID` via the `operand_elements` loop, there's no way to distinguish an explicit user annotation from one injected by this propagation. Raising 3-4 unconditionally was a false positive risk. Fixed in PR #85, motivated by issue #84. 
7. This is intencional, before this change, a Mixed operand raised a `Exception,` which ended as `UNKNOWN` in the validation report (issue #80, PR #78). Instead of that, returning `Mixed()` keeps the column unresolved rather than crashing and the type is resolved correctly when row data is available. On the SemanticError suggestion i agree, I will replace the bare Exception path with a proper SemanticError.
8. This is intencional, as DPM-XL spec §3.1.6, says: "Ranges (e.g., r0010-0100) select all Headers whose codes fall between the start and end codes, inclusive on both boundaries. The ordering is lexicographic". The old `_check_ranges_values_are_present` was contradicting the spec, causing false positive `1-2` errors on valid expressions (6 confirmed cases in PR #92).
9. Wasnt changed in this PR, already existed
10. Wasnt changed in this PR, already existed
11. Wasnt changed in this PR, already existed
12. Wasnt changed in this PR, already existed
13. This is fair, ill fix it
14. This is fair, ill fix it
15. Fair. Will add a `--quiet` flag

PR #105, opened 2026-06-03
feat: support APT-packaged SQLAlchemy 1.4.50 (dual 1.4/2.0)

### LINKED ISSUE #104: Support APT-packaged SQLAlchemy 1.4.50 (dual 1.4/2.0 compatibility)
## Summary

dpmcore requires `sqlalchemy>=2.0.49,<3.0` (locked 2.0.50) and its ORM is written entirely in SQLAlchemy 2.0 typed-declarative style. The target install platform — **Ubuntu 24.04 (Noble)** — ships `python3-sqlalchemy 1.4.50+ds1-1build1` via APT. The Noble CI job currently works around this with `pip install --break-system-packages "sqlalchemy>=2.0,<3.0"`, which defeats the goal of installing dpmcore cleanly from distro packages.

**Goal:** lower the SQLAlchemy floor to `>=1.4.50,<3.0` and make the ORM run on **both 1.4.50 and 2.0.x** (2.0 stays the canonical/typed target for local dev and the published wheel). The Noble CI job then installs `python3-sqlalchemy` from APT with no pip override.

## Scope

Three SQLAlchemy 2.0 features the ORM relies on are absent/different in 1.4 and must be fixed (required on both versions):

1. **Relationship targets** — all ~291 `relationship()` calls in `src/dpmcore/orm/*.py` have no positional target; the target lives only in the `Mapped["X"]` annotation, which 1.4 does not read. Add an explicit target string to each (format-preserving LibCST codemod).
2. **NOT NULL inference** — `nullable=` is used 0 times; non-Optional `Mapped[int]` infers NOT NULL on 2.0 but not on classic 1.4. Only **2** non-Optional, non-PK columns affected (`infrastructure.py` `action_id` ×2) — make them explicit `nullable=False`.
3. **Top-level `Engine`/`Connection`** — 2.0-only import location; move to `from sqlalchemy.engine import ...` in 7 files.

Plus: a `src/dpmcore/orm/_compat.py` shim (alias `mapped_column`→`Column`, expose `Mapped`/base on 1.4), `base.py` `DeclarativeBase`→`declarative_base()` rework keeping `to_dict`, `pyproject.toml` floor change + `poetry.lock` regen, and the CI workflow change below.

**Confirmed clean (no work):** no `declared_attr`, `hybrid_property`, `association_proxy`, `secondary=`, association tables, `server_default`/`default=`/`onupdate`, `TypeDecorator`, `__mapper_args__`, inheritance, `backref`, or 2.0-only column types. `select()`/`session.scalars()` and `session.query()` both work on 1.4.24+.

**Out of scope:** full classic rewrite / stripping `Mapped[]` annotations (fallback only if the 1.4.50 spike shows annotations are rejected); query-layer changes; Django models (no SQLAlchemy).

## Testing

- **Spike (GATE 0):** confirm 1.4.50 tolerates `Mapped[...]` annotations next to classic column values under PEP 563 (`from __future__ import annotations`).
- **Imports (GATE 1):** `import dpmcore.orm` + `configure_mappers()` clean on a 1.4.50 env.
- **Schema drift (GATE 2):** new `tests/unit/orm/test_schema_equivalence.py` comparing per-table metadata (type, nullable, PK, FK, unique/index) under both 1.4.50 and 2.0.50; manual `CreateTable` DDL diff compiled against the **postgresql + mssql** dialects (SQLite hides nullability/length drift; prod targets are PostgreSQL + SQL Server).
- **Behavior (GATE 3):** full `pytest tests/` green on both 2.0.50 (regression guard) and 1.4.50.
- **CI (GATE 4):** the Ubuntu 24.04 workflow green importing APT 1.4.x with the pip override removed.

## Additional Context

- APT package: `python3-sqlalchemy 1.4.50+ds1-1build1` (noble/main).
- Decision: **dual support** (`>=1.4.50,<3.0`), 2.0 remains canonical.
- The CI workflow already documents this tension: `.github/workflows/ubuntu_test_24_04.yml` ("Noble ships 1.4; dpmcore ORM requires 2.0 APIs").
- Note: the CI workflow change must land **with** the ORM compatibility work — removing the pip override before the ORM is 1.4-ready will fail the Noble job.

javihern98:
## Summary

Ubuntu 24.04 (Noble) ships `python3-sqlalchemy` **1.4.50** via apt, but dpmcore's ORM required 2.0-only APIs, forcing the Noble CI job to `pip install --break-system-packages "sqlalchemy>=2.0,<3.0"`. This makes the ORM run unchanged on **both 1.4.50 and 2.0.x**, lowers the floor to `>=1.4.50,<3.0` (2.0 stays the canonical/typed target), and switches the Noble CI to the apt package with no pip override.

Closes #104.

Key changes:
- **`src/dpmcore/orm/_compat.py`** (new) — version shim: aliases `mapped_column`→`Column` and re-exports `Mapped` on 1.4; `TYPE_CHECKING`-guarded so mypy always sees the 2.0 types.
- **`base.py`** — declarative base built per version (`DeclarativeBase` on 2.0, `declarative_base(cls=...)` on 1.4); engines/sessions use `future=True` so 1.4 gets 2.0-style `Connection`/`Session` semantics; `to_dict` uses `object_mapper()`.
- **~291 `relationship()` calls** got explicit target strings (LibCST codemod) — 1.4 does not infer targets from `Mapped[...]` annotations.
- **2 `action_id` columns** made explicit `nullable=False` (the only annotation-inferred NOT NULL).
- **7 files** import `Engine`/`Connection` from `sqlalchemy.engine` (valid on 1.4 + 2.0).
- **`structure.py`** — replaced `Row._tuple()` (2.0-only) with index access.
- **`ubuntu_test_24_04.yml`** — install `python3-sqlalchemy` via apt; drop the pip override.
- **`tests/unit/orm/test_schema_equivalence.py`** (new) — a dialect-aware schema snapshot (compiled against PostgreSQL + SQL Server) proving 1.4 and 2.0 emit an identical schema.

## Checklist

- [x] Code quality checks pass (`ruff format`, `ruff check`, `mypy --strict`)
- [x] Tests pass (`pytest`) — see Notes re: coverage gate
- [ ] Documentation updated (n/a — internal compatibility change)

## Impact / Risk

- **Breaking changes?** None. No public API / CLI / REST / Django model changes. ORM model behaviour is unchanged.
- **Database schema / migration concerns?** None — the emitted schema is **byte-identical** across 1.4.50 and 2.0.50 (68 tables; column types vs PostgreSQL+SQL Server, nullability, PK/FK/unique/index), verified by the new `test_schema_equivalence.py` and a one-off DDL diff.
- **Release/changelog:** dpmcore now supports SQLAlchemy 1.4.50 (Noble apt) alongside 2.0.x; dependency floor lowered to `>=1.4.50,<3.0`.

## Notes

- **Verification:** full suite green on **2.0.50** (1243 passed, 107 skipped) and on the **Noble-faithful 1.4.50 stack** — sqlalchemy 1.4.50 + pandas 2.1.4 + numpy 1.26.4 (1243 passed, 107 skipped). `mypy --strict` clean on 99 files.
- **pandas gotcha:** the *latest* pandas (2.2+/2.3) dropped SQLAlchemy-1.4 `to_sql` support; Noble's apt **pandas 2.1.4** works. Don't raise the pandas floor without re-checking the 1.4 `to_sql` path. (The Noble CI installs the wheel `--no-dependencies`, so it uses apt pandas 2.1.4.)
- **Coverage:** the 100% branch-coverage gate is currently disabled in `testing.yml` (`--cov-fail-under=0`, pending the `py_dpm` port). Version-specific shim branches carry `# pragma: no cover` so the gate stays green when re-enabled.
- `poetry.lock` only changes its content-hash (SQLAlchemy stays locked at 2.0.50 for the dev/wheel path).

### DIFF
--- a/.github/workflows/ubuntu_test_24_04.yml
+++ b/.github/workflows/ubuntu_test_24_04.yml
@@ -28,6 +28,10 @@ jobs:
         # httpx 0.26 removed that kwarg, so the combo is broken out of
         # the box. Installing fastapi via pip pulls a modern starlette
         # (>=0.37) that uses httpx.ASGITransport instead.
+        #
+        # python3-sqlalchemy is Noble's 1.4.50; dpmcore is compatible with
+        # both 1.4 and 2.0, so SQLAlchemy comes from apt with no pip
+        # override (see issue #104).
         run: |
           sudo apt-get update
           sudo apt-get install -y \
@@ -42,19 +46,21 @@ jobs:
             python3-pyodbc \
             python3-pytest \
             python3-rich \
+            python3-sqlalchemy \
             python3-uvicorn \
             unixodbc-dev
 
       - name: Install pip-only dependencies
-        # SQLAlchemy: Noble ships 1.4; dpmcore ORM requires 2.0 APIs.
         # fastapi: see the apt step above — pip pulls a starlette that
         # is compatible with apt httpx 0.26.
         # pytest-django: Noble 4.5.2 is below the dpmcore floor (>=4.9).
         run: |
           pip install --break-system-packages \
-            "sqlalchemy>=2.0,<3.0" \
             "fastapi>=0.101" \
             "pytest-django>=4.9"
+
+      - name: Show SQLAlchemy version (apt python3-sqlalchemy)
+        run: python3 -c "import sqlalchemy; print(sqlalchemy.__version__)"
 
       - name: Install poetry
         run: |
--- a/poetry.lock
+++ b/poetry.lock
@@ -3251,4 +3251,4 @@ sqlserver = ["pyodbc"]
 [metadata]
 lock-version = "2.1"
 python-versions = ">=3.10,<4.0"
-content-hash = "ccb62475de9b05c2275058fdb70ee87311ce8e913badbf3619092fc868f6d10b"
+content-hash = "7a201e7c65c6c0d86cbaf2c7511db50f643710efb42a7e7c0ba85a110e6df8a5"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -28,7 +28,10 @@ classifiers = [
     "Typing :: Typed",
 ]
 dependencies = [
-    "sqlalchemy>=2.0.49,<3.0",
+    # Floor is 1.4.50 to match the python3-sqlalchemy apt package on
+    # Ubuntu 24.04 (Noble); the ORM is compatible with both 1.4 and 2.0
+    # (see issue #104). 2.0 remains the canonical target for dev/wheel.
+    "sqlalchemy>=1.4.50,<3.0",
     # Runtime must match the ANTLR version used to generate the grammar
     # in src/dpmcore/dpm_xl/grammar/generated/. Regenerating with a newer
     # ANTLR is tracked separately.
--- a/src/dpmcore/connection.py
+++ b/src/dpmcore/connection.py
@@ -19,7 +19,8 @@ from __future__ import annotations
 from types import TracebackType
 from typing import TYPE_CHECKING, Any, Dict, Optional
 
-from sqlalchemy import Engine, create_engine
+from sqlalchemy import create_engine
+from sqlalchemy.engine import Engine
 from sqlalchemy.orm import Session, sessionmaker
 
 # Ensure all ORM modules are imported so relationships resolve.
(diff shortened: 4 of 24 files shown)


javihern98:
Addressed the `github-code-quality` feedback on `tests/unit/orm/test_schema_equivalence.py`: dropped the redundant `import dpmcore.orm` line. Verified that `from dpmcore.orm import Base` alone executes `dpmcore/orm/__init__.py` and registers all 68 tables on `Base.metadata`, so the schema snapshot is still complete (test re-run: green). Folded into the existing commit.

guillermo-garcia-1 (approved):

PR #118, opened 2026-06-08
Feat/analytic operators

### LINKED ISSUE #111: feat(dpm-xl): implement analytical (windowing) invocation for aggregate operators
### Initial Checks

### Problem Description

dpmcore cannot parse or evaluate aggregate operators with an `over` clause, nor the `rank` operator. These are needed to compute aggregates over windows of related Records without collapsing them

Spec: https://gitlab.com/aollerosr/dpm-xl-docs/-/work_items/88
Grammar commit: https://gitlab.com/aollerosr/dpm-xl-docs/-/commit/1bb77fe

Scope:
- Lexer: new ANALYTIC_CLAUSE_MODE; tokens RANK, OVER, PARTITION_BY, ORDER_BY,
  DATA_POINTS, CURRENT_DATA_POINT, RANGE, BETWEEN, PRECEDING, FOLLOWING,
  UNBOUNDED, ASC, DESC
- Parser: extend commonAggrOp to (groupingClause | analyticClause)?; add
  rankOp; add rules analyticClause, partitionClause, orderClause, orderItem,
  windowClause, windowBoundary
- AST: new node types for analytic invocation and rank
- Visitor / evaluator: window-function semantics; data points vs range frames
- Tests: partition only, order only, data points frame, range frame, rank,
  ties in ordering values

### Proposed Solution

```python
-- aggregate with over (sum, count, min_aggr, max_aggr, avg, median):
op(expression over(
  [partition by keyNames {, keyNames}]
  [order by keyNames [asc|desc] {, keyNames [asc|desc]}]
  [{data points | range} between boundary and boundary]
))

-- rank:
rank(expression over(
  [partition by keyNames {, keyNames}]
  order by keyNames [asc|desc] {, keyNames [asc|desc]}
))

-- window boundaries:
unbounded preceding | n preceding | current data point | n following | unbounded following

-- over and group by are mutually exclusive on the same call
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

ruizmaa:
## Summary

Closes #111 

Implements analytical (windowing) invocation for aggregate operators, as specified in dpm-xl-docs commit `1bb77fe`. 

Aggregate operators (`sum`, `count`, `avg`, `median`, `min_aggr`, `max_aggr`) previously only supported reducing invocation (`group by`), which collapses the RecordSet. This PR adds an `over(...)` clause that computes the aggregate over a window of related records while preserving the full RecordSet structure.

It also introduces the `rank` operator, which is always analytical and always returns an `Integer` fact.

## Checklist

## Impact / Risk


andres-sole (changes requested):
Some things i've found:

### Ranking by the value itself doesn't work. 
The whole point of rank is usually to rank rows by their measure: `rank({vRS} over (order by f desc)`. The parser accepts this, and there's even a test that confirms it parses. But when validation actually runs, it throws an error saying f isn't a valid component. That's because the check only looks at the key columns (like r, c) and treats the fact/measure column f as unknown. So the most common use of rank is broken, and no test caught it because every semantic test orders by a key, never by f. Is this intentional?

Suggested: allow ordering by the fact column, and add a test.

### There's a method that's never used.
Rank.validate_analytic exists but nothing ever calls it: rank actually runs through a different method. Dead code.

Suggested: Delete

### Two different ways of doing the same thing.
The normal aggregates (sum, avg, …) call one method for windowing, but rank does it through a differently-shaped method that doesn't match the parent class (it needs a type-ignore to compile).

Suggested: make rank go through validate_analytic like the other windowing operators (it's already the natural home for this logic), and drop the validate override and its type-ignore. That makes all windowing go through one consistent path and removes the dead Rank.validate_analytic method from point 2 at the same time.

andres-sole (approved):

Issue #120, opened 2026-06-08
Script generation


antonio-olleros:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/dpmcore/issues?q=is%3Aissue) for a similar question
- [x] I have checked the [documentation](https://docs.dpmcore.meaningfuldata.eu)

### Question

For these validations:

COREP_OF-4.1.0, IF_CLASS2-1.4.0 -> (v09808_m, v22973_m-v22977_m)
IF_CLASS2-1.4.0, IF_CLASS3-1.4.0 -> (v11120_m, v11121_m)
COREP_OF-3.1.0, IF_CLASS2-1.2.0 -> (v0643_m, v0644_m, v0645_m)
COREP_OF-4.1.0, IF_CLASS2-1.3.0 -> (v0580_m, v0624_m, v0629_m, v0640_m)

Should be included in the COREP-OF script as without dependencies.

### Context

```python

```

antonio-olleros:
For v0580_m

```
with {default: 0, interval: true}: {tC_18.00, r0010, c0070, s0001} = {tC_02.00.a, r0540, c0010}
```

The scopes yield a conflict for COREP_OF 4.0.0, because it is intra-module and cross-module (because for IF there is only one of the two tables, so for running IF, would require COREP_OF. 

When generating the scripts, we should take this into account as follows:
-  If there are intra-instance and cross-instance options, choose intra-instance
- If only cross, take cross

This would result in COREP_OF having this validations as intra-instance, and IF_CLASS2 as cross-instance

PR #123, opened 2026-06-09
Fix: scope calculation errors

### LINKED ISSUE #119: Check scopes
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/dpmcore/issues?q=is%3Aissue) for a similar question
- [x] I have checked the [documentation](https://docs.dpmcore.meaningfuldata.eu)

### Question

Please check the scopes of these validations:

COREP_OF-4.1.0, IF_CLASS2-1.4.0 -> (v09808_m, v22973_m-v22977_m)
IF_CLASS2-1.4.0, IF_CLASS3-1.4.0 -> (v11120_m, v11121_m)
COREP_OF-3.1.0, IF_CLASS2-1.2.0 -> (v0643_m, v0644_m, v0645_m)
COREP_OF-4.1.0, IF_CLASS2-1.3.0 -> (v0580_m, v0624_m, v0629_m, v0640_m)

They should yield only intra-instance scopes.

### Context

```python

```

### dpmcore version

_No response_

### Additional Context

_No response_

andres-sole:
Closes #119 #120 

## Checklist

## Overview
**1. Filing-indicator preconditions were never matched.**
`model_queries.py` filtered preconditions with `Variable.type == "Filing Indicator"` (spaced + capitalised). As a result they never counted toward a module's coverage, so a module that really did host the whole operation was not seen as intra-instance.

**2. Preconditions were miscounted (NaN bucket).**
The target count was `len(table_codes) + len(precondition_items)`, but the per-module check only counted `TableCode.unique()`. Precondition rows keep their code in the `Code` column and are NaN in `TableCode`, so all preconditions collapsed into one NaN bucket. With one or more preconditions the counts never matched and a full module was wrongly pushed into the cross-module pool.

**3. Cross-instance scopes were incomplete or spurious.**
- Partial modules (holding only some of the referenced tables) were combined directly, without first pulling in modules that provide the tables they are missing → incomplete "cross" scopes.
- Modules from non-overlapping lifecycle generations (different reference-date windows) were combined even though they are never reported together.
- Invalid combinations were still emitted, just with `from_submission_date = None`, instead of being dropped.

**4. Intra/cross conflict in script generation (#120).**
When the primary module could evaluate the operation on its own (intra) AND also showed up in cross scopes (because a sibling module needs it), the operation was classified ambiguously. For scripts, the primary's script should treat it as intra; only the sibling's script should treat it as a cross-instance dependency.

### v0580_m  (#119 and #120)

```
with {default: 0, interval: true}:
{tC_18.00, r0010, c0070, s0001} = {tC_02.00.a, r0540, c0010}
```

**Before (reported in #120):** conflict — COREP_OF came out as BOTH intra-module and cross-module at the same time (because IF has only one of the two tables, so running it under IF would require COREP_OF). The script generation had no rule to resolve the conflict.

**After:**

```
total_scopes=2  is_cross_module=True
  Scope 1 [INTRA] from=2026-03-31: COREP_OF-4.1.0
  Scope 2 [CROSS] from=2026-03-31: COREP_OF-4.1.0, IF_CLASS2-1.4.0
```

For script generation the prefer-intra rule then gives:
- COREP_OF → **intra-instance** (no dependency)
- IF_CLASS2 → **cross-instance** dependency on COREP_OF

### v09808_m  (#119)

```
with {default:0, interval:true}:
{tC_34.02.a, (c0020..c0220), (r0010), s*} + ... + {tC_34.02.a, ..., (r0030), s*}
+ {tC_34.02.b, ..., (r0040), s*} + ... + {tC_34.02.b, ..., (r0100), s*}
= {tC_34.02.b, ..., (r0110), s*}
```

**Before:** the precondition/counting bug plus the incomplete cross logic produced spurious cross-instance scopes instead of clean intra scopes.

**After:**

```
total_scopes=2  is_cross_module=False
  Scope 1 [INTRA] from=2026-03-31: IF_CLASS2-1.4.0
  Scope 2 [INTRA] from=2026-03-31: COREP_OF-4.1.0
```

### DIFF
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -64,6 +64,23 @@ if TYPE_CHECKING:
 # Helper utilities
 # ------------------------------------------------------------------ #
 
+# The DPM 2.0 Refit schema stores the filing-indicator variable type as
+# the single token "filingindicator". Some source exports spell it
+# "Filing Indicator" (with a space and capitals), so matching is done on
+# a case- and whitespace-normalised form rather than a fixed literal.
+_FILING_INDICATOR_TYPE = "filingindicator"
+
+
+def _is_filing_indicator() -> Any:
+    """Return a SQLAlchemy clause matching filing-indicator variables.
+
+    Normalises ``Variable.type`` (lower-cased, spaces removed) before
+    comparison so the match is robust to spelling variants across DPM
+    source exports (``"Filing Indicator"`` vs ``"filingindicator"``).
+    """
+    normalized = func.lower(func.replace(Variable.type, " ", ""))
+    return normalized == _FILING_INDICATOR_TYPE
+
 
 def _get_engine_cache_key(session: "Session") -> Hashable:
     """Return a hashable key that identifies the engine.
@@ -343,8 +360,8 @@ class VariableVersionQuery:
         """Find a filing-indicator variable by code.
 
         Looks for a VariableVersion whose code matches
-        *variable_code* and whose Variable type is
-        ``'Filing Indicator'``.
+        *variable_code* and whose Variable type is a
+        filing indicator (see :func:`_is_filing_indicator`).
 
         Args:
             session: SQLAlchemy session.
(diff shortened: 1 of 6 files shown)


antonio-olleros (approved):
Looks great

ruizmaa (approved):

PR #161, opened 2026-06-26
Support custom code and version_id in preconditions API

### LINKED ISSUE #160: ASTGeneratorService.script preconditions API can't carry the precondition's code/version_id, breaking key parity
### Initial Checks

### Summary

The preconditions parameter of ASTGeneratorService.script accepts only (expression, [validation_codes]) per precondition — no way to pass the precondition's own Operation.Code or OperationVersion.OperationVID. dpmcore therefore generates keys like P_<variable_vid> (e.g. P_395473) while the legacy MDPM tool uses P_<operationID> (e.g. P_571). The precondition AST content is byte-identical between both tools; only the naming differs. Affects 73 of 79 successfully-generated modules in the parity audit — the most universal divergence.

### Reproducible Example

```python
On branch mdpm-parity-investigation:
poetry run python scripts/generate_module_from_db.py IRRBB 1.2.0 --out /tmp/IRRBB.json
poetry run python -c '
import json
out = next(iter(json.load(open("/tmp/IRRBB.json")).values()))
print("dpmcore keys :", sorted(out["preconditions"]))
print("MDPM equivalent: P_564, P_565, P_566, P_567, P_568, P_569, P_570, P_571, P_572")
'
```

### Expected vs Actual

Callers that know the precondition's identity should be able to pass it through. Suggested signature (backward-compatible):

preconditions = [
    {"expression": "(v_)08.00)", "affected_operations": ["v16317_s"],
     "code": "P_571", "version_id": 8341},
    ...
]

# or tuple form: (expression, [validation_codes], code=None, version_id=None)

When metadata is absent, fall back to the current variable_vid convention. The code, version_id, and dictionary key of each precondition entry should use the supplied values when provided.

Actual behavior:
dpmcore keys: ['P_395473', 'P_395474', 'P_395475', 'P_395476', 'P_395477', 'P_395480', 'P_395483', 'P_395486', 'P_395489']
All keys, code fields, and version_id values are hard-coded to derive from variable_vid at src/dpmcore/services/ast_generator.py:757 (key = f'P_{vi["variable_vid"]}') and :765. The caller cannot override.

### dpmcore version

0.1.1rc3

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

Affects 73/79 modules — the most universal divergence with MDPM. The AST (PreconditionItem with variable_id/variable_code) is correct; only naming is wrong, so this is a low-risk additive change to the API.

ruizmaa:
Closes #160 

## Summary

The preconditions API generates keys based on `variable_vid` (e.g., `P_395473`), while MDPM uses the precondition's own `operation_id` (e.g., `P_571`). This naming divergence affects 73 of 79 modules. There was no way for callers to specify the correct code and version_id.

So the fix is to allow callers to optionally provide `code` and `version_id` in precondition specifications. The implementation accepts both tuple format (old, backward-compatible) and dict format (new, with custom fields). Updated REST API, service layer, and CLI to handle both formats. When custom values are provided, they override auto-generated defaults.

## Checklist

### DIFF
--- a/src/dpmcore/cli/main.py
+++ b/src/dpmcore/cli/main.py
@@ -13,6 +13,7 @@ Usage::
 from __future__ import annotations
 
 import sys
+from typing import Any
 
 import click
 
@@ -419,6 +420,30 @@ def serve(database: str, host: str, port: int) -> None:
     type=click.Path(dir_okay=False),
     help="Path to write the generated script JSON.",
 )
+def _process_preconditions_from_json(
+    preconditions_raw: list[Any],
+) -> list[tuple[str, list[str]] | dict[str, Any]] | None:
+    """Convert raw precondition entries to script() format."""
+    if not preconditions_raw:
+        return None
+
+    preconditions: list[tuple[str, list[str]] | dict[str, Any]] = []
+    for entry in preconditions_raw:
+        has_custom = "code" in entry or "version_id" in entry
+        if isinstance(entry, dict) and has_custom:
+            preconditions.append(entry)
+        elif isinstance(entry, (list, tuple)) and len(entry) >= 2:
+            preconditions.append((entry[0], list(entry[1])))
+        elif isinstance(entry, dict) and "expression" in entry:
+            preconditions.append(
+                (
+                    entry["expression"],
+                    list(entry.get("affected_operations", [])),
+                )
+            )
+    return preconditions or None
+
+
 def generate_script(
     expressions_path: str,
     module_code: str,
@@ -472,7 +497,7 @@ def generate_script(
 
     items = [tuple(pair) for pair in raw["expressions"]]
     preconditions_raw = raw.get("preconditions") or []
-    preconditions = [(pair[0], list(pair[1])) for pair in preconditions_raw]
+    preconditions = _process_preconditions_from_json(preconditions_raw)
 
     severities_raw = raw.get("severities")
     severities: dict[str, str] | None = None
--- a/src/dpmcore/cli/main.py
+++ b/src/dpmcore/cli/main.py
@@ -367,6 +367,30 @@ def serve(database: str, host: str, port: int) -> None:
     uvicorn.run(app, host=host, port=port)
 
 
+def _process_preconditions_from_json(
+    preconditions_raw: list[Any],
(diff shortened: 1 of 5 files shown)


src/dpmcore/cli/main.py
```diff
@@ -419,6 +420,30 @@ def serve(database: str, host: str, port: int) -> None:
     type=click.Path(dir_okay=False),
     help="Path to write the generated script JSON.",
 )
+def _process_preconditions_from_json(
```
andres-sole: _process_preconditions_from_json is defined between the @main.command/@click.option decorators and 
  def generate_script, so the entire generate-script command is bound to the wrong function.
  All of @main.command("generate-script") + the @click.option(...) stack (lines 370–421) now decorate _process_preconditions_from_json
  (which takes only preconditions_raw), while def generate_script at line 447 is left undecorated and orphaned — never registered,
  never called. Running dpmcore generate-script --expressions … --module-code … --output … raises TypeError: 
  _process_preconditions_from_json() got an unexpected keyword argument 'expressions_path' and exits non-zero (reproduced).

src/dpmcore/services/ast_generator.py
```diff
@@ -755,29 +778,38 @@ def _collect_precondition_var_infos(
...
+        When provided_code or provided_version_id are supplied, they override
+        the auto-generated values.
         """
         if len(var_infos) == 1:
             vi = var_infos[0]
-            key = f"p_{vi['variable_vid']}"
-            return key, {
+            default_key = f"p_{vi['variable_vid']}"
+            code = provided_code or default_key
+            version_id = provided_version_id or vi["variable_vid"]
```
andres-sole: or-based override treats a falsy version_id of 0 as "not provided".
  version_id = provided_version_id or vi["variable_vid"] (and the compound case … or sorted_vids[0]) silently discards a
  caller-supplied version_id of 0, falling back to the auto-generated vid. provided_code or default_key has the same issue for an
  empty-string code. version_id 0 is implausible for real DPM data, but the intent is "override when supplied" — use
  provided_version_id if provided_version_id is not None else

andres-sole (changes requested):

andres-sole (approved):

PR #163, opened 2026-06-26
Skip invalid operations instead of aborting the whole module

### LINKED ISSUE #155: ASTGeneratorService.script aborts entire module when a single expression references a grey or non-existent cell
### Initial Checks

### Summary

When a single expression in a batch references a grey cell or a non-existent cell, ASTGeneratorService.script returns success: False and discards the whole module — every other valid validation is lost. In a parity audit across 107 modules, ≥9 modules failed this way (AE-1.2.0, COREP_ALM-3.1.0, CODIS-2.0.0, and 6 others). The wholesale abort happens at src/dpmcore/services/ast_generator.py:182 inside the for item in expressions loop.

### Reproducible Example

```python
On branch mdpm-parity-investigation with a DPM 4.2.1 SQLite DB at dpm_4.2.1_20260624.db:
poetry run python scripts/generate_module_from_db.py AE 1.2.0 --out /tmp/AE-1.2.0.json
The helper loads all AE-1.2.0 validations from the DB (380+ ops) and feeds them to ASTGeneratorService.script. See scripts/generate_module_from_db.py:_load_module for the exact SQL.
```

### Expected vs Actual

Expected: The result should be {"success": True, "enriched_ast": {...}} with the offending ops listed in a failed_operations block (or simply skipped) — every other op should appear in the output.

Actual Behavior: Script generation failed: Grey cells {F_32.03.a, r0040, c0010}, {F_32.03.a, r0050, c0010}, {F_32.03.a, r0060, c0010} were found.
enriched_ast is None. All 380+ ops in the module are lost, even ones that would have validated cleanly.

### dpmcore version

0.1.1rc3

### Python version

3.12.3

### Database backend

SQLite (dpm_4.2.1_20260624.db, DPM release 4.2.1)

### OS

Ubuntu 24.04.4 LTS

### Additional Context

Same wholesale-abort pattern hits a different error class (This operation is not allowed, because there is no match between DPM components) in 5 other modules — see Issue ASTGeneratorService.script aborts module with "No match between DMP components". Both should be fixable by the same per-op resilience change. Full list of affected modules in scripts/parity_report.json on the same branch.

ruizmaa:
Closes #155 

## Summary

When `ASTGeneratorService.script` encountered a semantic validation failure on any expression (grey cells, missing cells, incompatible DPM components), it aborted the entire module and returned `success: False` with no output.

The fix changes the per-expression validation check from a module-level abort into a skip: invalid operations are recorded in a new `failed_operations` dict (`{code: error_message}`) and the loop continues. The successful return now includes this dict alongside the existing `enriched_ast`, so callers can inspect which operations were dropped and why

## Checklist

### DIFF
--- a/src/dpmcore/server/routers/scripts.py
+++ b/src/dpmcore/server/routers/scripts.py
@@ -55,6 +55,7 @@ class GenerateScriptResponse(BaseModel):
     success: bool
     enriched_ast: Optional[Any] = None
     error: Optional[str] = None
+    failed_operations: Dict[str, str] = {}
 
 
 def create_scripts_router(
@@ -99,6 +100,7 @@ def create_scripts_router(
             success=bool(result.get("success")),
             enriched_ast=result.get("enriched_ast"),
             error=result.get("error"),
+            failed_operations=result.get("failed_operations", {}),
         )
 
     return router
--- a/README.md
+++ b/README.md
@@ -125,10 +125,12 @@ with connect("sqlite:///dpm.db") as db:
     #                    dpm_release, dates, operations, variables, tables,
     #                    preconditions, precondition_variables,
     #                    dependency_information, dependency_modules}}
+    # script["failed_operations"] lists expressions that were skipped due to
+    # semantic errors (e.g. grey cells): {validation_code: error_message}
     namespace, ns_block = next(iter(script["enriched_ast"].items()))
-    print(ns_block["dpm_release"])      # {"release": "...", "publication_date": "..."}
-    print(ns_block["operations"])       # {validation_code: {ast, severity, ...}}
-    print(ns_block["dependency_modules"])
+    print(ns_block["dpm_release"])          # {"release": "...", "publication_date": "..."}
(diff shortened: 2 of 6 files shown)


src/dpmcore/server/routers/scripts.py
```diff
@@ -55,6 +55,7 @@ class GenerateScriptResponse(BaseModel):
     success: bool
     enriched_ast: Optional[Any] = None
     error: Optional[str] = None
+    failed_operations: Dict[str, str] = {}
```
andres-sole, line 58: failed_operations is only in the success return. 5 of the 6 return paths in script() (no-session, precondition ValueError, scope error, except ValueError, except Exception) omit it, i think it should appear too

ruizmaa: Already fixed with [b996b01](https://github.com/Meaningful-Data/dpmcore/pull/163/commits/b996b01a7f97d00409185a62a06933cab0abdc05)

andres-sole (approved):

PR #164, opened 2026-06-29
Calculation dependency graph

andres-sole:
## Checklist

## Impact / Risk


ruizmaa (dismissed):

tests/unit/services/test_calculations_graph.py
```diff
@@ -0,0 +1,283 @@
...
+explicit-reference behaviour against a *minimal* in-memory database (a single
+release, no cell views). Exact cell resolution against real data is covered by
+``tests/integration/services/test_calculations_graph_resolution.py``.
+"""
+
+import pytest
+from sqlalchemy import create_engine
+from sqlalchemy.orm import Session
+
+import dpmcore.orm  # noqa: F401 — register ORM mappers
```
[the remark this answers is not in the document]
andres-sole: deliberate side-effect import: it forces every ORM submodule to load so SQLAlchemy cross-module relationships resolve before any query run

ruizmaa (dismissed):

src/dpmcore/services/calculations_graph/engine.py
```diff
@@ -0,0 +1,299 @@
...
+            "Invalid release code",
+            f"Release {release_code!r} has no ordering.",
+        )
+    kept: list[tuple[Any, ...]] = []
+    for row in rows:
+        start_so = sort_orders.get(row[3])
+        if start_so is None or start_so > target:
+            continue
+        end_so = sort_orders.get(row[4]) if row[4] is not None else None
+        if end_so is not None and end_so < target:
```
guillermo-garcia-1: With target release 3, a row ending at release 3 is kept here — it should be excluded, like a row ending at release 2. The check only excludes end < target; it should exclude end <= target (require end > target, or no end at all). This is the half-open convention the shared filter_by_release helper uses everywhere else in the codebase, including for OperationVersion itself in every other query — this is the one place that does it differently.

src/dpmcore/services/calculations_graph/service.py
```diff
@@ -0,0 +1,626 @@
...
+        if not header or header[0].strip().lower() != "code":
+            raise Invalid(
+                "Invalid calculations script",
+                f"Expected a 'Code,Expression' header, got: {header!r}.",
+            )
+        for row in reader:
+            if not row or not row[0].strip():
+                continue
+            code = row[0].strip()
+            expression = row[1].strip() if len(row) > 1 else ""
```
guillermo-garcia-1: _read_csv_rows only keeps row[0] and row[1]. DPM-XL expressions almost always contain commas, so an unquoted CSV row splits into more than 2 fields — everything past the second is silently dropped, and the truncated expression fails to parse with no clear error.

src/dpmcore/services/calculations_graph/engine.py
andres-sole: I've just confirmed it. Fixed by deleting the row filter and applying `filter_by_release` to the query. Tests updated also.

src/dpmcore/services/calculations_graph/service.py
andres-sole: Fixed. Now everything after the code is rejoined into the expression. Added details to docs and a test covering it.

guillermo-garcia-1 (approved):

PR #183, opened 2026-07-01
Fix standalone persistent assignment with cell-reference LHS

### LINKED ISSUE #181: Persistent assignment ({cellRef} <- expr) rejected with "unexpected '<-'"
### Initial Checks

### Summary

A persistent assignment expression with a cell reference on the left-hand side (`{tK_61.00, r0010, c0010} <- {tC_01.00, r0020, c0010}`) is rejected by the parser with `Unexpected '<-' - expecting end of expression or ';'`.

The same operator works correctly when nested inside a temporary assignment (`v1204 := {vFoo} <- {tC_01.00, r0020, c0010}`).

### Reproducible Example

```python

svc = SyntaxService()

# Works
result = svc.validate("v1204 := {vFoo} <- {tC_01.00, r0020, c0010}")
print(result.is_valid)  # True

# Fails — should be valid
result = svc.validate("{tK_61.00, r0010, c0010} <- {tC_01.00, r0020, c0010}")
print(result.is_valid)   # False
print(result.error_message)  # mismatched input '<-' expecting {<EOF>, ';'}
```

### Expected vs Actual

Expected: both expressions parse as valid. A cell reference is a legitimate target for a persistent assignment.

Actual: the standalone form raises a parse error: `mismatched input '<-' expecting {<EOF>, ';'}`

### dpmcore version

0.1.1

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #181 

## Summary

A persistent assignment with a cell reference on the left-hand side (`{tK_61.00, r0010, c0010} <- {tC_01.00, r0020, c0010}` was rejected by the parser with mismatched input `'<-' expecting {<EOF>, ';'}`)

`persistentAssignmentExpression` was never reachable as a top-level statement, and its LHS only accepted variable references (`varID`), not cell references. The fix adds it to `statement` and widens the LHS to select, which covers both.

## Checklist

## Impact / Risk


tests/unit/ast/test_persistent_assignment.py
```diff
@@ -0,0 +1,62 @@
...
+assignments.
+"""
+
+import pytest
+
+from dpmcore.dpm_xl.ast.nodes import PersistentAssignment, VarID, VarRef
+from dpmcore.services.syntax import SyntaxService
+
+# Syntax validity
+VALID_PERSISTENT_ASSIGNMENTS = [
```
andres-sole, line 14: Right now an assignment to a parameter ( {p_x, number} <- 1 ) would pass but makes no sense. Changing the grammar is cumbersome but i thought it was worth flagging out

ruizmaa: In any case, this could be restricted at the semantic level rather than the grammar level if that's simpler.

andres-sole (dismissed):

tests/unit/ast/test_persistent_assignment.py
ruizmaa: I gave it a try at the grammar level and it wasn't that cumbersome. Just a new `assignmentTarget` rule restricting the LHS to `cellRef` and `varRef`. Happy to include it in this PR if it makes sense to you:

```
persistentAssignmentExpression:
    assignmentTarget PERSISTENT_ASSIGN expressionWithoutAssignment
    ;

assignmentTarget:
    CURLY_BRACKET_LEFT (cellRef | varRef) CURLY_BRACKET_RIGHT
    ;
```

src/dpmcore/dpm_xl/grammar/dpm_xlParser.g4
```diff
@@ -37,7 +38,11 @@ temporaryAssignmentExpression:
     ;
 
 persistentAssignmentExpression:
-    varID PERSISTENT_ASSIGN expressionWithoutAssignment
+    assignmentTarget PERSISTENT_ASSIGN expressionWithoutAssignment
```
andres-sole, line 41: Best solution to the problem

andres-sole (approved):

PR #186, opened 2026-07-03
Switch release ordering logic from semver to date based

### LINKED ISSUE #185: Release ordering fails when the release code is not semver
### Initial Checks

### Summary

Release ordering is computed by parsing the release code as `MAJOR.MINOR[.PATCH]`. Any release whose code is not in that shape cannot be ordered, which breaks release-filtered queries.

### Reproducible Example

```python
from dpmcore import connect

# A module that only exists in a working / "Playground" release,
# whose release code is not MAJOR.MINOR[.PATCH].
    db.services.layout_exporter.export_module(
        "CODIS", release_code="9999"
    )
```

### Expected vs Actual

Expected: the export runs and returns the module's tables for that release.

Actual: it crashes with
`ValueError: release <id> has no sort_order — its code could not be parsed as MAJOR.MINOR[.PATCH].`

### dpmcore version

0.1.1

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Closes #185

Release ordering now runs off `Release.date` instead of parsing the version code, so every code format orders the same way — classic `MAJOR.MINOR[.PATCH]`, EBA's four-segment codes like `4.2.1.3`, and non-versioned working releases like `"Playground"` — without crashing on codes that aren't semver. A release with no publication date is an unpublished working release and is treated as the latest, so requesting it returns the current rows. This rule is applied consistently everywhere releases are ordered: range/scope filtering, `get_last_release`, the release listing and single-release lookups, the structure "latest" endpoint, and the export filename. `get_last_release` and the "latest" lookups now pick by date rather than by the opaque `ReleaseID`, which is no longer monotonic from DPM 4.2.1 onwards.

## Checklist

## Impact / Risk

- No API, CLI, REST, or Django model changes.
- No schema or migration changes
- Behaviour change: an undated release now ranks as the latest instead of being excluded, and release ordering follows publication date rather than the version code.

### DIFF
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -952,30 +952,29 @@ class ModuleVersionQuery:
     def get_last_release(
         session: "Session",
     ) -> int | None:
-        """Return the ID of the most recent release, ordered by date.
+        """Return the ID of the latest release by the date sort order.
 
-        Releases are ordered by ``Release.date`` (descending), not the
-        opaque ``release_id`` FK, which is non-monotonic from DPM 4.2.1
-        onwards (e.g. ``4.2.1`` is ``ReleaseID 1010000003``), so the
-        highest id is not necessarily the latest release. Releases with
-        no date are unrankable and excluded.
+        Releases are ranked by :func:`compute_sort_order` (from
+        ``Release.date``), not the opaque ``release_id`` FK, which is
+        non-monotonic from DPM 4.2.1 onwards (e.g. ``4.2.1`` is
+        ``ReleaseID 1010000003``), so the highest id is not necessarily
+        the latest release. An undated (unpublished) release sorts as the
+        latest; ties are broken by ``release_id`` for determinism.
 
         Args:
             session: SQLAlchemy session.
 
         Returns:
-            Integer release ID of the latest dated release, or ``None``
-            when there is no dated release.
+            Integer release ID of the latest release, or ``None`` when
+            there are no releases.
         """
-        result = (
-            session.query(Release.release_id)
-            .filter(Release.date.isnot(None))
-            .order_by(Release.date.desc())
-            .first()
-        )
-        if result is None:
+        sort_orders = load_release_sort_orders(session)
+        ranked = [
+            (so, rid) for rid, so in sort_orders.items() if so is not None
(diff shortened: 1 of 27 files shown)


victorprietosandamaso-cloud:
Verified locally on cr-185. Three scenarios covered: non-semver code ('Playground'), undated release ranking as latest ('Working'), and four-segment code ('4.2.1.3'). All pass after the last fixes. Evidence attached in issue #185

victorprietosandamaso-cloud (dismissed):

src/dpmcore/orm/release_sort_order.py
ruizmaa: This can never happen

src/dpmcore/orm/release_sort_order.py
```diff
@@ -1,95 +1,80 @@
...
-    parsed = parse_version(code)
-    if parsed is None:
-        return None
-    major, minor, patch = parsed
-    return (
-        major * _SEGMENT_BITS * _SEGMENT_BITS + minor * _SEGMENT_BITS + patch
-    )
+    if release_date is None:
+        return _UNDATED_SORT_ORDER
+    return release_date.toordinal()
```
ruizmaa, line 55: `compute_sort_order` never returns `None` anymore, that leaves dead code in:

- `ModuleVersionQuery.get_last_release` (`model_queries.py`): filters `if so is not None` over `sort_orders.items()`, but that, branch can no happen.
- `release_ids_for_sort_order` (`release_sort_order.py`) has `if so is None: continue` inside the loop

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -952,18 +952,29 @@ class ModuleVersionQuery:
...
         Returns:
-            Integer release ID, or None.
+            Integer release ID of the latest release, or ``None`` when
+            there are no releases.
         """
-        result = session.query(func.max(Release.release_id)).scalar()
-        if result is None:
+        sort_orders = load_release_sort_orders(session)
+        ranked = [
+            (so, rid) for rid, so in sort_orders.items() if so is not None
```
ruizmaa: `if so is not None` branch can never happen

ruizmaa (approved):

PR #189, opened 2026-07-08
feat(services): add ExpressionMetadataService and data-dictionary gap…

### LINKED ISSUE #188: bug(expression_metadata): get_referenced_headers drops axes when same header code is used on multiple axes in one table
### Initial Checks

### Summary

ExpressionMetadataService.get_referenced_headers emits at most one row per (table_vid, header_code). If the same header code is used on more than one axis inside a single table version (for example row and column with the same code), only the first axis in the internal ordering is returned; the other axis is silently dropped. This makes the result incomplete for callers that persist the full set of referenced headers.

The cause is the break statement after the first non-seen (code, usage, table_vid) key inside the for usage in ordered: loop in
src/dpmcore/services/expression_metadata.py (around lines 255–271).  need every axis that the expression actually uses.

### Reproducible Example

```python
from dpmcore import connect

# Assumes a release where table T has a header code that is used
# on both the row axis and the column axis inside the same
# table_vid (this can happen on tables with symmetric layouts).
    svc = db.services.expression_metadata
    expr = "{tT, r0010, c0010} = 100"

    headers = svc.get_referenced_headers(
        expression=expr, release_code="4.2.1"
    )

    # Filter to the shared code.
    same_code = [h for h in headers if h["code"] == "0010"]
    axes = {h["header_type"] for h in same_code}
    print(axes)
```

### Expected vs Actual

Expected: {"Row", "Column"} for the affected table_vid — every axis
          the expression uses must appear in the result.

Actual:   Only one of {"Row", "Column"} appears for that table_vid;
          the other axis is missing.

### dpmcore version

0.1.1

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

- Suggested fix: remove the break inside for usage in ordered: and keep the seen set to deduplicate; the loop will then emit one row per (table_vid, code, usage).
- If the singular-per-table behavior is really desired, document it explicitly in the docstring and add a fixture-backed test that
exercises the multi-axis case so the choice is visible. 
- Related: unit tests for the ordering fallback branch (catalog usage not in usages) are missing, so the current behavior is not covered by the branch coverage gate.

### LINKED ISSUE #190: feat(services): expose tables/headers/frameworks referenced by a DPM-XL expression + data-dictionary gaps
### Initial Checks

### Problem Description

Callers that persist a validation expression (for example a Django app saving a `ValidationVersion`) need to know which tables, headers and frameworks the expression touches. Today the only way to get that is to reach into `ScopeCalculatorService` internals or re-parse the expression by hand. Both options leak ORM instances into the caller and duplicate parsing.

The data-dictionary service also has two small gaps compared to the old pydpm tool:

- `get_tables` returns only codes, so callers must issue a second query to get the name and description.
- There is no public way to fetch the open-key (compound-key) variables of a table; the logic lives in a private helper on `ScopeCalculatorService`.

### Proposed Solution

```python
from dpmcore import connect

    meta = db.services.expression_metadata
    expr = "{tF_01.01, r0010, c0010} = 100"

    tables = meta.get_referenced_tables(expr, release_code="4.2.1")
    headers = meta.get_referenced_headers(expr, release_code="4.2.1")
    frameworks = meta.get_referenced_frameworks(expr, release_code="4.2.1")

    dd = db.services.data_dictionary
    verbose = dd.get_tables(release_code="4.2.1", verbose=True)
    keys = dd.get_open_keys_for_table("F_01.01", release_code="4.2.1")
```

### Alternatives Considered

- Extend `ScopeCalculatorService` directly. Rejected: that service is focused on module-version scope; mixing entity resolution into it would blur its purpose.
- Return ORM instances. Rejected: callers would need to keep the session open and would leak ORM types across layers.

### Additional Context

Anchored on the canonical case `{tF_01.01, r0010, c0010} = 100` at release `4.2.1` (both FINREP9 and FINREP9DP resolve). Motivated by the adam-ide migration from pydpm to dpmcore.

victorprietosandamaso-cloud:
## Summary

- New `ExpressionMetadataService` returns the tables, headers and frameworks referenced by a DPM-XL expression as plain dicts. No ORM instances leak to callers. Syntax and semantic errors return `[]`.
- `DataDictionaryService.get_tables(..., verbose=True)` returns `{code, name, description}` per row.
- `DataDictionaryService.get_open_keys_for_table(s)` exposes the open-key variables of a table.
- `db.services.expression_metadata` is wired on the connection accessor (lazy, cached).
- Docs: new `docs/api/expression_metadata.rst`, README example, index entry.

Closes #188
Closes #190

## Checklist

- [ ] Tests pass (`pytest`) with 100% branch coverage — integration tests skip without fixture DB; unit tests for defensive branches still needed
- [x] Documentation updated

## Impact / Risk

- Public API additions only.
- `get_tables` return type widened from `List[str]` to `List[Any]` to accept the `verbose=True` shape. Consider `typing.overload` in a follow-up.
- `get_open_keys_for_tables` currently calls a private method on `Sc-up to promote it to a shared helper.
- No DB schema or migration changes.

## Notes

- See #188 for the known axis-drop bug in `get_referenced_headers`.
- Unit tests for defensive branches still to be added to keep branch

### DIFF
--- a/src/dpmcore/services/expression_metadata.py
+++ b/src/dpmcore/services/expression_metadata.py
@@ -28,7 +28,7 @@ service holds no state beyond its SQLAlchemy session.
 from __future__ import annotations
 
 import logging
-from typing import TYPE_CHECKING, Any, Dict, List, Optional
+from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple
 
 from sqlalchemy import and_, or_
 
@@ -99,17 +99,14 @@ class ExpressionMetadataService:
         On a syntax/semantic error (or when the expression references
         nothing that resolves), returns ``[]``.
         """
-        oc = self._prepare_operands(expression, release_id, release_code)
-        if oc is None:
+        prepared = self._prepare_operands(expression, release_id, release_code)
+        if prepared is None:
             return []
+        oc, release_id = prepared
 
         table_vids = self._extract_table_vids(oc)
         if not table_vids:
             return []
-
-        release_id = resolve_release_id(
-            self.session, release_id=release_id, release_code=release_code
-        )
 
         query = (
             self.session.query(
@@ -194,8 +191,11 @@ class ExpressionMetadataService:
 
         On a syntax/semantic error, returns ``[]``.
         """
-        oc = self._prepare_operands(expression, release_id, release_code)
-        if oc is None or oc.data is None or oc.data.empty:
+        prepared = self._prepare_operands(expression, release_id, release_code)
+        if prepared is None:
+            return []
+        oc, _ = prepared
+        if oc.data is None or oc.data.empty:
             return []
 
         table_vids = self._extract_table_vids(oc)
@@ -295,17 +295,14 @@ class ExpressionMetadataService:
 
         On a syntax/semantic error, returns ``[]``.
         """
-        oc = self._prepare_operands(expression, release_id, release_code)
-        if oc is None:
+        prepared = self._prepare_operands(expression, release_id, release_code)
+        if prepared is None:
             return []
+        oc, release_id = prepared
 
         table_vids = self._extract_table_vids(oc)
         if not table_vids:
             return []
-
-        release_id = resolve_release_id(
-            self.session, release_id=release_id, release_code=release_code
-        )
 
         query = (
             self.session.query(Framework)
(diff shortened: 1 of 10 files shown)


src/dpmcore/services/expression_metadata.py
```diff
@@ -0,0 +1,418 @@
...
+if TYPE_CHECKING:
+    from sqlalchemy.orm import Session
+
+logger = logging.getLogger(__name__)
+
+
+# Mapping from DPM ``Header.direction`` to the header-usage label the
+# expression syntax uses ("r*"→Row, "c*"→Column, "s*"→Sheet). Kept in
+# sync with the label emitted by :meth:`get_referenced_headers`.
+_DIRECTION_TO_USAGE = {"X": "Row", "Y": "Column", "Z": "Sheet"}
```
andres-sole: The X/Y axes seem to be inverted inverted here. should be {"X": "Column", "Y": "Row", "Z": "Sheet"}. Please confirm but the rest of the codebase should align.

src/dpmcore/services/expression_metadata.py
```diff
@@ -0,0 +1,418 @@
...
+                        "header_vid": row.header_vid,
+                        "code": row.code or "",
+                        "label": row.label or "",
+                        "header_type": usage,
+                        "table_vid": row.table_vid,
+                        "table_code": row.table_code or "",
+                        "table_name": row.table_name or "",
+                    }
+                )
+                break
```
andres-sole: This break emits at most one per DB row, which drops the second axis when a code is used on bothh

tests/integration/services/test_expression_metadata.py
```diff
@@ -0,0 +1,142 @@
...
+
+    headers = service.get_referenced_headers(
+        expression=_EXPRESSION, release_id=release_id
+    )
+
+    assert headers, "expected header rows"
+
+    # r0010 must be exposed as Row-usage; c0010 as Column-usage.
+    row_labels = {h["header_type"] for h in headers if h["code"] == "0010"}
+    assert {"Row", "Column"}.issubset(row_labels)
```
andres-sole, line 81: inverted _DIRECTION_TO_USAGE mapping still passes. tighten

andres-sole (approved):

PR #193, opened 2026-07-09
feat(dpm_xl): first-class set operators (union/intersect/setdiff/symdiff, empty set, set equality)

### LINKED ISSUE #192: Update DPM-XL grammar and semantic layer to align with docs MR !74 (set operators)
### Summary

Upstream docs repository (dpm-xl-docs) MR !74 (https://gitlab.com/aollerosr/dpm-xl-docs/-/merge_requests/74) reshapes the set-operator area of the grammar and its semantic rules. dpmcore currently mirrors the pre-MR grammar; expressions that rely on the new rules parse but fail semantic analysis.

End users need expressions such as
`setdiff(union({tX, cA}[get qINC], {tY, cB}[get qINC]), {tZ, cC}[get qINC]) = {}`
to validate against dpmcore. Today they fail.

### Scope

AST constructor (`src/dpmcore/dpm_xl/ast/constructor.py`):
- `visitUnion/Intersect/Setdiff/SymdiffSetExpr`: children are `ExpressionContext`.
- `visitSetOperand`: handle empty `{}`.
- Dispatch for the new `#setExpr` alternative.
- Drop dead visitors: `visitCountSetOp`, `visitSubcategorySelectExpr`.

Semantic analyzer (`src/dpmcore/dpm_xl/semantic_analyzer.py`):
- `visit_Set`: support empty `Set`.
- `_visit_set_operands`: implicit `set_of` coercion on RecordSet operands (§13.1.5).

Operators (`src/dpmcore/dpm_xl/operators/base.py`):
- `types_given_structures`: accept `ScalarSet == ScalarSet` and `ScalarSet == RecordSet`.
- `validate_types`: scalar path accepts `ScalarSet = ScalarSet` as valid set-equality (§13.7).

Out of scope: script-generation/execution layer for set equality (follow-up).

### Testing

- Add unit tests for: empty set literal `{}`, `union`/`setdiff`/`symdiff` with RecordSet operands coerced via implicit `set_of`, set equality `= {}` and `!= {}`.
- Verify existing set-operator tests still pass.
- Manual: full pipeline validation of
  `setdiff(union({tC_28.00, c020}[get qINC], {tC_29.00, c030}[get qINC]), {tC_27.00, c021}[get qINC]) = {}`
  passes both syntax and semantic services.
- Preserve 100% branch coverage; ruff/mypy clean

### Additional Context

Upstream MR: https://gitlab.com/aollerosr/dpm-xl-docs/-/merge_requests/74

victorprietosandamaso-cloud:
## Summary

Expands DPM-XL's set-operator support so that set-valued expressions are
first-class throughout the grammar, semantic layer, and script
generation pipeline. Previously the grammar accepted a narrow set of
shapes, the semantic layer refused several legitimate compositions, and
the script generator dropped set operators silently.

### Grammar / AST

- `union` / `intersect` / `setdiff` / `symdiff` now take generic
  `expression` operands, so they compose freely with clause operators
  (e.g. `union({tT, c010}[get qINC], {tT2, c020}[get qINC])`).
- The empty set literal `{}` is now valid syntax.
- `in`'s right-hand side and `count`'s operand are generic expressions.
- `setExpression` is a first-class alternative of `expression`.

### Semantic layer

- Recordset operands to set operators are implicitly coerced to their
  Fact Component's ScalarSet (equivalent to wrapping in `set_of`).
- Set equality (`=` / `!=`) between two ScalarSets returns a Boolean
  Scalar. Mixing a ScalarSet with a Recordset on `=` / `!=` is rejected
  in both directions.
- `count` over a ScalarSet returns an Integer Scalar (via
  `AggregationOp`, replacing the previous dedicated code path).
- Empty set operands are excluded from the operand-type homogeneity
  check so `union({}, {1,2})` and friends validate cleanly.
- Arithmetic / ordering / boolean operators over ScalarSets, and
  ill-shaped operand combinations on the scalar path, now raise a
  precise `SemanticError("3-3")` instead of leaking a bare
  `NotImplementedError` or `Exception` (which showed up as
  `error_code="UNKNOWN"`).
- `in` with a non-set RHS (e.g. `5 in 3`) is rejected at semantic time
  with `SemanticError("3-3")` instead of failing later with a runtime
  `TypeError`.

### Script generation

- `ASTToJSONVisitor` gains visitors for `Set`, `SetOfOp`, `UnionSetOp`,
  `IntersectSetOp`, `SetdiffOp`, `SymdiffOp`. These nodes previously
  had no visitor and were silently dropped from the enriched AST
  payload, so any script using a set operator came out empty.
- `MLGeneration` replaces the placeholder `NotImplementedError` stubs
  for the same six operators with real operation-node emissions that
  follow the existing `BinOp` / `ComplexNumericOp` patterns.

### Tests

Unit tests added across:
- `tests/unit/semantic/test_set_operators.py`
- `tests/unit/dpm_xl/test_operator_invariants.py`
- `tests/unit/ast/test_set_operators.py`
- `tests/unit/ast/test_ml_generation.py`

Covering: empty-set handling, `count` over ScalarSet, set equality on
ScalarSet pairs, rejection of mixed ScalarSet/Recordset in equality,
non-set RHS in `in`, invalid operand combinations on the scalar path,
JSON serialization of all new AST nodes (including nested), and the
`MLGeneration` visitors emitting the correct operation nodes.

Fixes #192

## Impact / Risk

- No public API changes.
- ANTLR outputs regenerated with 4.9.2 (matches the pinned runtime).
- No DB schema change.
- Behavioural change: set operators now accept Recordset operands
  (implicit `set_of`), `=` / `!=` between ScalarSets are valid, set
  operators are emitted into the generated script (previously raised
  `NotImplementedError` in `MLGeneration`), and several previously
  untyped `Exception` / `NotImplementedError` failures are now typed
  `SemanticError("3-3")`.


src/dpmcore/dpm_xl/operators/base.py
```diff
@@ -553,6 +558,11 @@ def validate(
         # ``accepts_scalar_set_rhs`` (currently only the ``in`` membership
         # operator). Any other shape — notably a bare RecordSet leaking past
         # validate_structures — is a real bug and gets a precise rejection.
+        if isinstance(left, ScalarSet) and isinstance(right, ScalarSet):
```
andres-sole: the `ScalarSet == ScalarSet` branch lives in the base `Operator.validate`, so every binary operator inherits it, not just `=`/`!=`. `{1,2} + {3,4}`, `{1,2} * {3,4}`, `{1,2} > {3,4}` all pass semantic validation. Gate it to `Equal`/`NotEqual`

src/dpmcore/dpm_xl/grammar/dpm_xlParser.g4
```diff
@@ -55,7 +55,7 @@ expression:
     | left=expression op=(PLUS|MINUS) right=expression                                                  #numericExpr
     | left=expression op=CONCAT right=expression                                                        #concatExpr
     | left=expression op=comparisonOperators right=expression                                           #compExpr
-    | left=expression op=IN setExpression                                                               #inExpr
+    | left=expression op=IN right=expression                                                            #inExpr
```
andres-sole, line 58: Widening the `in` RHS from `setExpression` to `expression` lets a scalar RHS through with no set-guard. `5 in 3` now parses and dies with a bare `TypeError: argument of type 'int' is not iterable`, not a `SemanticError`. Reject a non-set RHS in `In.validate` / `visit_BinOp`

ruizmaa:
Checked the semantic implications of relaxing the grammar and found this gaps:

1. `count` over a set is broken: `count({1,2,3})` now raises `SemanticError("4-4-0-1")` ("Only a Recordset is allowed..."). Dropping `#countSetOp` routes it through `visit_AggregationOp`, which requires `RecordSet` unconditionally. Needs a `ScalarSet` branch there.
2. Bare `NotImplementedError` leaks on invalid set expressions: `{1,2,3} + {1,2}`, `{1,2,3} > {1,2}` now parse but crash with an unhandled `NotImplementedError` in `types_given_structures`, with an empty `error_code="UNKNOWN"`. Should raise `SemanticError("3-3", ...)` instead, like `In` already does.
3. `{}` breaks union/intersect/setdiff/symdiff: `union({}, {1,2})` fails with "types are wrong, expected homogeneous scalar type" (`Item` vs `Integer`). The homogeneity check in `_visit_set_operands` compares every operand's type as if it were a real element type, but `{}` has no elements to infer a type from, so it's given an arbitrary placeholder type (`Item`), which then looks like a type conflict against any other set. An empty set can't actually conflict with anything type-wise, so it should be skipped by that check.
4. Set equality (`=`/`!=`) silently accepts a `ScalarSet` mixed with a `RecordSet` in either direction (e.g. `setdiff(A,B) = {c020}` -> `Scalar`, `{c020} = setdiff(A,B)` -> `RecordSet`), when §13.7.5/§5.2.1.5 explicitly require this to be rejected. It doesn't crash or error, it returns a valid result but should raise a `SemanticError`

```python
import pandas as pd
import dpmcore.dpm_xl.semantic_analyzer
from dpmcore.dpm_xl.semantic_analyzer import InputAnalyzer
from dpmcore.dpm_xl.operators.comparison import Equal
from dpmcore.dpm_xl.symbols import FactComponent, KeyComponent, RecordSet, ScalarSet, Structure
from dpmcore.dpm_xl.types.scalar import Integer
from dpmcore.dpm_xl.utils.tokens import STANDARD

# 1:
InputAnalyzer("x").visit(SyntaxService().parse("count({1,2,3})"))
# SemanticError: Only a Recordset is allowed in the Aggregation operator count.

# 2:
InputAnalyzer("x").visit(SyntaxService().parse("{1,2,3} + {1,2}"))
# NotImplementedError

# 3:
InputAnalyzer("x").visit(SyntaxService().parse("union({}, {1,2})"))
# SemanticError: Types are wrong, type_op_1=Item, Integer, operator works for homogeneous scalar type.

# 4:
ss = ScalarSet(type_=Integer(), name=None, origin="{1,2,3}")
rs = RecordSet(Structure([KeyComponent("r", Integer(), STANDARD, "t"), FactComponent(Integer(), "t")]), "t", "t")
rs.records = pd.DataFrame({"r": ["1"], "data_type": [Integer()]})
type(Equal.validate(ss, rs)).__name__, type(Equal.validate(rs, ss)).__name__
# ('Scalar', 'RecordSet')  # both should raise SemanticError
```

src/dpmcore/dpm_xl/operators/base.py
```diff
@@ -267,8 +269,22 @@ def types_given_structures(
             return fact_component.type, right.type, op_type_to_check
         elif isinstance(left, Scalar) and isinstance(right, ScalarSet):
```
andres-sole, line 270: scalar = set-literal leaks a bare Exception instead of a semantic error

andres-sole (approved):

andres-sole (approved):

PR #195, opened 2026-07-09
fix(scopes): taxonomy url point to the right release

### LINKED ISSUE #194: Taxonomy URL uses the report release instead of the release where the module version was published
### Initial Checks

### Summary

When dpmcore builds the taxonomy URL for a module, it uses the release you pass in the request. It should instead use the release where that module version was first published.

These two releases are not always the same. A point release (for example `4.2.1`) changes only a few modules. Those modules get a new version and are published under `4.2.1`. Every other module stays exactly the same as in the previous release (`4.2`) and is still published under `4.2`.

For a module that did not change, dpmcore still writes the report release (`4.2.1`) into its URL. But no taxonomy for that module exists at `4.2.1`, so the URL points to something that is not there. This breaks cross-module dependencies, because the dependency URL of an unchanged module does not match the URL where its taxonomy actually lives.

### Reproducible Example

```python
from dpmcore import connect

# Cross-module rule: F_32.01 is in module AE, F_01.01 is in module FINREP9/FINREP9DP.
EXPR = "sum ({tF_32.01, r0010, (c0010, c0060)}) = {tF_01.01, r0380, c0010}"

    result = db.services.ast_generator.script(
        expressions=[(EXPR, "rule1")],
        module_code="FINREP9DP",   # a module that DID change in 4.2.1 (version 1.1.0)
        module_version="1.1.0",
        release="4.2.1",           # the report release
    )

import json
text = json.dumps(result["enriched_ast"])

# AE did not change in 4.2.1, so its taxonomy is published under 4.2.
print("ae/4.2.1/mod/ae" in text)   # prints True  -> the bug (this URL does not exist)
print("ae/4.2/mod/ae" in text)     # prints False -> this is the URL that should be used
```

### Expected vs Actual

Expected: the AE dependency URL is
  http://www.eba.europa.eu/eu/fr/xbrl/crr/fws/ae/4.2/mod/ae
  (release 4.2, where AE version 1.4.0 was published)

Actual: the AE dependency URL is
  http://www.eba.europa.eu/eu/fr/xbrl/crr/fws/ae/4.2.1/mod/ae
  (release 4.2.1, which has no AE taxonomy)

### dpmcore version

0.1.2rc1

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Closes #194 

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/services/scope_calculator.py
+++ b/src/dpmcore/services/scope_calculator.py
@@ -652,24 +652,19 @@ class ScopeCalculatorService:
     ) -> Optional[str]:
         """Resolve a module VID to its EBA taxonomy URI.
 
-        Resolution order:
+        The URI's release segment always comes from the release in which
+        the module version was introduced (its start release), resolved
+        via :meth:`_resolve_uri_release_id`. A module version's taxonomy
+        is published under that release, so an unchanged module keeps its
+        original release segment even inside a later report: e.g. an
+        unchanged ``ae`` stays at ``.../ae/4.2/mod/ae`` in a 4.2.1 report,
+        because no ``ae`` taxonomy exists at 4.2.1. The ``release_id``
+        argument only selects which module version is active; it does not
+        set the release segment.
 
-        - When ``release_id`` is supplied (the script-generation path),
-          build the URI dynamically using *that* release's code as the
-          release segment, regardless of when the module version
-          itself was introduced. This matches what the EBA taxonomy
-          package writers do: every module shipped inside a given
-          taxonomy release is rooted at ``.../{release}/mod/...``,
-          even when the underlying module version's start release is
-          older. Sharing the release segment across all modules in a
-          script is what lets the engine resolve cross-instance
-          dependencies against the matching XBRL Report Packages.
-
-        - When ``release_id`` is not supplied (ad-hoc lookups outside
-          script generation), try the static CSV mapping by
-          ``(module_code, version_number)`` first; on a miss, fall
-          back to the dynamic builder using the module version's
-          start release.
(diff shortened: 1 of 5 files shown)


src/dpmcore/services/scope_calculator.py
ruizmaa: Argument `release_id` is now never used in `_get_module_uri()`. Also `_build_dependency_entry` and `_map_pairs_to_uris` are still passing it. Should this be removed?

src/dpmcore/services/scope_calculator.py
```diff
@@ -652,24 +652,19 @@ def _get_module_uri(
...
-          start release.
+        The URI's release segment always comes from the release in which
+        the module version was introduced (its start release), resolved
+        via :meth:`_resolve_uri_release_id`. A module version's taxonomy
+        is published under that release, so an unchanged module keeps its
+        original release segment even inside a later report: e.g. an
+        unchanged ``ae`` stays at ``.../ae/4.2/mod/ae`` in a 4.2.1 report,
+        because no ``ae`` taxonomy exists at 4.2.1. The ``release_id``
+        argument only selects which module version is active; it does not
+        set the release segment.
```
ruizmaa: `release_id` doesn't select the module version either, the mv lookup only filters by `module_vid`, no `release_id` filter exists in the function

andres-sole: Removal done

src/dpmcore/services/scope_calculator.py
andres-sole: Removed in latest commit

ruizmaa (approved):

PR #198, opened 2026-07-09
fix(semantic): validate range endpoints exist in cell selections

### LINKED ISSUE #196: Range endpoints are not validated for existence in semantic validation
### Initial Checks

### Summary

Semantic validation does not check that both endpoints of a range selection actually exist. A range with a non-existent endpoint is silently accepted instead of being reported as an error.

In a range selection the first character is the selector type (`c` for column, `r` for row, ...) and the rest of that first token is the code of the first endpoint. After the `-`, only the code of the second endpoint is written; the selector type is not repeated.

So `c0010-0030` means "columns from 0010 to 0030". If a user mistakenly repeats the `c` and writes `c0010-c0030`, the range is read as "from code `0010` to code `c0030`". In table F_11.03 the column code `0030` exists but `c0030` does not, so no range can be resolved and the expression should fail semantically.

Existence of a referenced code is already checked for the table selector (`F_11.03` must resolve). The same check should apply to the endpoints of a range, for every selector that can carry one (columns, rows, ...) — not just to the table code. Today the range endpoints are not checked, so this mistake passes validation.

### Reproducible Example

```python
from dpmcore import connect

    # Intended by the writer (valid): a column range 0010 -> 0030
    intended = db.services.semantic.validate(
        "with {tF_11.03, c0010-0030, default:0}: "
        "{r0010} >= {r0020} + {r0030} + {r0040}",
        release_code="4.2.1",
    )
    print("intended:", intended.is_valid)  # True (correct)

    # Typo: a second "c" in the range endpoint.
    # Parsed as the range from code "0010" to code "c0030".
    # Column "c0030" does not exist in F_11.03 ("0030" does).
    typo = db.services.semantic.validate(
        "with {tF_11.03, c0010-c0030, default:0}: "
        "{r0010} >= {r0020} + {r0030} + {r0040}",
        release_code="4.2.1",
    )
    print("typo:", typo.is_valid)  # currently True, should be False
```

### Expected vs Actual

Expected: the second expression fails semantic validation with a "missing column" error, reporting that column `c0030` does not exist in table F_11.03.

Actual: the second expression passes semantic validation. The second endpoint of the range is never checked for existence, so the typo goes unnoticed.

The requested fix is that both endpoints of a range are validated for existence in semantic validation, for every range-capable selector (columns, rows, ...), the same way the table code is already validated.

### dpmcore version

0.1.2rc1

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
# fix(semantic): validate range endpoints exist in cell selections

## Summary

Fixes #196.

Semantic validation did not check that both endpoints of a range selection actually exist. A range resolves through a `between` comparison, which silently accepts a bogus endpoint: the comparison still spans the real codes in between, so no cell is reported missing and the mistake passes validation unnoticed.

Concretely, the typo `c0010-c0030` (a repeated selector letter) is read as the range from code `0010` to code `c0030`. Column `c0030` does not exist in F_11.03, yet the expression was accepted as valid.

This change checks both endpoints of a range against the codes available for that selector before resolving it, for every range-capable selector (columns, rows, sheets). A missing endpoint raises the existing `1-2` "cell not found" error, echoing the offending code in selector notation — the same existence check already applied to the table code and to explicitly listed cells. The "cells not found" raise was extracted into a small helper so the range path and the listed-cell path report identically; the existing message is unchanged.

## Checklist

New unit tests pass locally; the DB-backed integration test and full-suite coverage run in CI. No documentation change needed.

## Notes

New tests: `tests/unit/dpm_xl/test_data_handlers.py`.

### DIFF
--- a/src/dpmcore/dpm_xl/utils/data_handlers.py
+++ b/src/dpmcore/dpm_xl/utils/data_handlers.py
@@ -59,32 +59,41 @@ def filter_data_by_cell_element(
     cell_elements: Sequence[str],
     element_name: str,
     table_code: str,
+    valid_codes: set[str] | None = None,
 ) -> pd.DataFrame:
-    """Filter data by cell elements
+    """Filter data by cell elements.
+
     :param series: data to be filtered
     :param cell_elements: rows, columns or sheets using to filter data
     :param element_name: name of cell elements using to filter data
+    :param table_code: table code, echoed in "cell not found" errors
+    :param valid_codes: the full set of codes defined for ``element_name``
+        in the table, used only to validate range endpoints. When ``None``
+        the codes present in ``series`` are used. Callers that narrow one
+        axis after another (see :func:`filter_all_data`) must pass the
+        table-wide set, so a range endpoint that is a real code but does
+        not intersect the already-selected cells is not misreported as a
+        missing cell.
     :return: filtered data.
     """
+    if valid_codes is None:
+        valid_codes = set(series[element_name])
     if len(cell_elements) == 1 and "-" not in cell_elements[0]:
         series = series[series[element_name] == cell_elements[0]]
     elif len(cell_elements) == 1 and "-" in cell_elements[0]:
         limits = cell_elements[0].split("-")
-        _check_range_endpoints(
-            set(series[element_name]), limits, element_name, table_code
-        )
+        _check_range_endpoints(valid_codes, limits, element_name, table_code)
         series = series[series[element_name].between(limits[0], limits[1])]
     else:
         range_control = any("-" in x for x in cell_elements)
         if range_control:  # Range in cell elements, we must separate them
-            available = set(series[element_name])
             data_range = []
             data_single = []
             for x in cell_elements:
                 if "-" in x:
                     limits = x.split("-")
                     _check_range_endpoints(
-                        available, limits, element_name, table_code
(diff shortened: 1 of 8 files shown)


andres-sole:
Endpoint check I added rejected some valid ranges on sparse tables. `filter_all_data` filters rows first, then columns. The check was validating column endpoints against the columns left after the row filter, not all columns in the table.

Example: `c0010-0030` on a row where `c0030` is grey. `c0030` is a real column, but it has no cell in that row, so the check reported it as missing and raised `1-2`. Before this branch the range just resolved.

 Fix: validate endpoints against the table's full set of codes (captured before any filtering). Filtering itself is unchanged. The #196 typo (`c0030` isn't a real column at all) still fails.

Range endpoints are now more lenient than listed cells: a listed grey `c0030` still errors, a range endpoint doesn't.

ruizmaa (commented):
Seems good to me, only concern is the same one I left on #195, since this branch shares its base and carries those commits too.

andres-sole:
> Seems good to me, only concern is the same one I left on #195, since this branch shares its base and carries those commits too.

Updated in base branch and included those commits here

ruizmaa (dismissed):

PR #206, opened 2026-07-13
fix(ast): downgrade deprecated "null" string literal to a warning

### LINKED ISSUE #205: Deprecated != "null" (string-literal) syntax raises SemanticError
### Initial Checks

### Summary

 Environment
  - dpmcore==0.1.1
  - Python 3.12

Deprecated != "null" (string-literal) syntax raises SemanticError from the AST visitor with message "Cannot use null literal, you must use the isnull function". These are historically valid expressions that shipped in older framework.

DB releases; a warning would be more appropriate than a hard error, since the grammar accepts the input and the visitor understands what the author meant.

### Reproducible Example

```python
from antlr4 import CommonTokenStream, InputStream
  from dpmcore.dpm_xl.grammar.generated.dpm_xlLexer import dpm_xlLexer
  from dpmcore.dpm_xl.grammar.generated.dpm_xlParser import dpm_xlParser
  from dpmcore.dpm_xl.ast.constructor import ASTVisitor

  expression = (
      'with {tB_05.01, default: 0, interval: false}: '
      'if({c0030} != "null") then ({c0040} != "null") endif'
  )
  
  lexer = dpm_xlLexer(InputStream(expression))
  parser = dpm_xlParser(CommonTokenStream(lexer))
  tree = parser.start()          # grammar accepts the input

  ASTVisitor().visit(tree)
  # raises dpmcore.errors.SemanticError:
  # "Cannot use null literal, you must use the isnull function"
```

### Expected vs Actual

Expected: raises dpmcore.errors.SemanticWarning
Current raises dpmcore.errors.SemanticError

### dpmcore version

0.1.1

### Python version

3.12

### Database backend

SQLite

### OS

Ubuntu 24.04

### Additional Context

_No response_

victorprietosandamaso-cloud:
## Summary

Historical DPM-XL expressions in dpmcore's supported DB releases use a
``"null"`` string literal (a STRING_LITERAL whose text is the word
``null``) as a null sentinel — for example:

if({c0030} != "null") then ({c0040} != "null") endif

The AST visitor aborted these with ``SemanticError("0-3")`` — a hard
failure — even though the grammar accepts the input and the author's
intent is unambiguous.

``visitLiteral`` now materialises the ``"null"`` string literal as
``Constant(type_="Null", value=None)`` (the same node the bare
``null`` keyword produces via ``default: null``) and surfaces a
deprecation warning through ``add_semantic_warning`` instead of
raising.

To make sure the warning reaches ``SemanticResult.warning``,
``SemanticService.validate`` now wraps ``_syntax.parse(...)`` inside
its ``collect_warnings`` context. Warnings emitted from AST
construction are captured alongside the ones from the analyzer pass.

Tests added in ``tests/unit/ast/test_default_null.py`` covering: no
exception on the reproducer expression, ``"null"`` literal
materialises as a ``Null`` Constant, warning emitted at parse time,
and both occurrences of ``"null"`` in the reproducer expression
contribute to the warning stream.

Fixes #205

## Checklist

## Impact / Risk

- No public API changes.
- No DB schema change.
- Behavioural change: expressions containing the ``"null"`` string
  literal now validate with a deprecation warning instead of failing
  with ``SemanticError("0-3")``. Downstream code that treated ``0-3``
  as an expected error for these inputs will see success with a
  warning instead.

## Notes

- Warning is surfaced through ``SemanticResult.warning``. Callers of
  the lower-level ``ASTVisitor().visit(...)`` outside a
  ``collect_warnings`` context still get the AST; the warning is
  silently discarded on that path.

### DIFF
--- a/src/dpmcore/dpm_xl/ast/constructor.py
+++ b/src/dpmcore/dpm_xl/ast/constructor.py
@@ -811,7 +811,7 @@ class ASTVisitor(dpm_xlParserVisitor):
                 # used the string literal ``"null"`` as a null sentinel. The
                 # grammar accepts it and the intent is unambiguous, so treat
                 # it as a proper Null Constant and surface a deprecation
-                # warning instead of aborting semantic analysis with 0-3.
+                # warning instead of aborting semantic analysis.
                 add_semantic_warning(
                     'Deprecated use of the "null" string literal; '
                     "prefer the ``isnull(...)`` function or the bare "
--- a/src/dpmcore/dpm_xl/ast/constructor.py
+++ b/src/dpmcore/dpm_xl/ast/constructor.py
@@ -64,6 +64,7 @@ from dpmcore.dpm_xl.grammar.generated.dpm_xlParserVisitor import (
     dpm_xlParserVisitor,
 )
 from dpmcore.dpm_xl.utils.tokens import TABLE_GROUP_PREFIX
+from dpmcore.dpm_xl.warning_collector import add_semantic_warning
 from dpmcore.errors import SemanticError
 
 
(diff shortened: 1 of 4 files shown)


ruizmaa (commented):
Should we delete the error message `"0-3"` as is no longer used anymore??

ruizmaa (approved):

PR #222, opened 2026-07-15
fix: bridge ghost siblings when window-checking explicit release

### LINKED ISSUE #221: Ghost-fallback not applied in ast_generator._resolve_explicit_release: fails with "past the end" when the explicit release falls outside the strict window of the non-ghost MV
### Initial Checks

### Summary

When a validation is executed with an explicit release and the ModuleVersion covering that release is a ghost (FromReferenceDate == ToReferenceDate), the ghost-fallback logic introduced in #182 (dpm_xl/model_queries.py::_resolve_with_ghost_fallback) correctly redirects content resolution to the previous non-ghost MV. However, services/ast_generator.py::_resolve_explicit_release (line 407) applies a strict window-check on that fallback MV and rejects the release because it falls outside the MV's original window. The two code paths are inconsistent: scope resolution knows about ghosts, explicit-release resolution does not.

### Reproducible Example

```python
Using the adam-ide DB, DORA has three module versions where v1.0.0 and v1.2.0 are ghosts.

from dpmcore import connect

  with connect("<db_url>") as db:
      # calculate_from_expression correctly falls back to v1.1.0 via ghost-fallback:
      r = db.services.scope_calculator.calculate_from_expression(
          expression='match({tB_01.01, c0020}[get ERI], "[A-Z0-9]{18}[0-9]{2}")',
          release_id=5,   # 4.2
      )
      # r.module_versions == [435]  → DORA 1.1.0

      # But calling script() with that same v1.1.0 and release=4.2 fails:
      r = db.services.ast_generator.script(
          module_code="DORA",
          module_version="1.1.0",
          expressions=[(
              'match({tB_01.01, c0020}[get ERI], "[A-Z0-9]{18}[0-9]{2}")',
              "BDI_DORA_001",
          )],
          release="4.2",
      )
      # r["error"] == "Release 4.2 is past the end of module version DORA 1.1.0 (ends at release_id=5)."

      # With the ghost version explicitly, it succeeds:
      r = db.services.ast_generator.script(
          module_code="DORA",
          module_version="1.2.0",
          expressions=[(
              'match({tB_01.01, c0020}[get ERI], "[A-Z0-9]{18}[0-9]{2}")',
              "BDI_DORA_001",
          )],
          release="4.2",
      )
      # r["success"] is True
```

### Expected vs Actual

Expected behavior

script(module_code="DORA", module_version="1.1.0", release="4.2") succeeds. Since v1.1.0 is the ghost-fallback for a release whose covering MV (v1.2.0) is a ghost,  _resolve_explicit_release should either skip the strict window-check for that MV or virtually extend the window to include the range originally covered by the ghost.

Actual behavior

 script(...) fails with: Release 4.2 is past the end of module version DORA 1.1.0 (ends at release_id=5). _resolve_explicit_release in services/ast_generator.py:447-454 applies the raw end_release_id window-check on v1.1.0 without any awareness of the ghost-fallback that upstream code just
  performed.

### dpmcore version

0.1.2rc2

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

victorprietosandamaso-cloud:
Body:
  ## Summary
  - `ast_generator._resolve_explicit_release` now consults `_effective_end_release_id`, which virtually extends the received `ModuleVersion`'s end past any adjacent chain of ghost siblings
  (same module, `FromReferenceDate == ToReferenceDate`) to the start of the next non-ghost sibling, or to `None` when only ghosts follow.
  - Aligns the strict window check with the ghost fallback in `dpm_xl.model_queries._resolve_with_ghost_fallback`, which already substitutes a ghost with its latest prior non-ghost. Without
  this fix, a validation issued with an explicit release just past the non-ghost fallback's end (e.g. DORA v1.1.0 ending at 4.2 with ghost v1.2.0 covering 4.2+) failed with `Release X is
  past the end of module version`. 
  - Predates branch is deliberately unchanged: `_latest_prior_non_collapsed_vids` picks strictly backward, so a legitimate ghost fallback can never produce that scenario.
  
  ## Test plan
  - [x] `pytest tests/integration/services/test_scope_release_axis.py` (18 passed — includes 6 new tests exercising: DORA-shape past-end, no-ghost negative, ghost of a different module,
  open-ended ghost, and extension terminating at the next non-ghost's `start_release_id`).
  - [x] `ruff format`, `ruff check`, `mypy src/` all clean.
  - [x] End-to-end reproduction on the DORA local DB: `ast_generator.script("DORA", "1.1.0", release="4.2")` now succeeds; the pre-fix `Release 4.2 is past the end of module version DORA
  1.1.0` no longer fires.

  Fixes #221

### DIFF
--- a/src/dpmcore/services/ast_generator.py
+++ b/src/dpmcore/services/ast_generator.py
@@ -52,6 +52,19 @@ def _normalize_variable_code(code: str) -> str:
     return code
 
 
+def _is_ghost_mv(mv: Any) -> bool:
+    """Whether a ``ModuleVersion`` is a ghost (collapsed reference window).
+
+    Mirrors :func:`_collapsed_mask` in ``dpm_xl.model_queries``: a ghost
+    has both reference dates set and equal.
+    """
+    return (
+        mv.from_reference_date is not None
+        and mv.to_reference_date is not None
+        and mv.from_reference_date == mv.to_reference_date
+    )
+
+
 def _format_date(value: Any, fallback: Optional[str] = None) -> Optional[str]:
     """Format a ``date`` / ``datetime`` / string as ``YYYY-MM-DD``."""
     if value is None:
@@ -420,6 +433,17 @@ class ASTGeneratorService:
         :mod:`dpmcore.orm.release_sort_order`), not the raw id. Raises
         ``ValueError`` if the release is unknown, predates
         ``start_release_id``, or is past ``end_release_id``.
+
+        Coordinates with the ghost-version fallback in
+        :func:`dpmcore.dpm_xl.model_queries._resolve_with_ghost_fallback`:
+        when ``mv`` is being used as the fallback for a ghost that
+        follows it, ``mv``'s effective end release is virtually extended
+        past the ghost(s) to the start of the next non-ghost sibling —
+        or ``None`` when only ghosts follow (issue #221). The
+        ``predates`` branch is never relaxed:
+        ``_latest_prior_non_collapsed_vids`` picks fallbacks strictly
+        backward, so a legitimate ghost-fallback can never produce an MV
+        whose start ``predates`` the requested release.
         """
         from dpmcore.orm.infrastructure import Release
         from dpmcore.orm.release_sort_order import resolve_sort_order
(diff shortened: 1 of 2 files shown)


src/dpmcore/services/ast_generator.py
```diff
@@ -52,6 +52,19 @@ def _normalize_variable_code(code: str) -> str:
...
+    Mirrors :func:`_collapsed_mask` in ``dpm_xl.model_queries``: a ghost
+    has both reference dates set and equal.
+    """
+    return (
+        mv.from_reference_date is not None
+        and mv.to_reference_date is not None
+        and mv.from_reference_date == mv.to_reference_date
+    )
+
+
```
ruizmaa: I don't think it's a good idea to have `_is_ghost_mv` as a separate function for just a couple of conditions when it's only used once

src/dpmcore/services/ast_generator.py
```diff
@@ -444,16 +468,147 @@ def _resolve_explicit_release(
...
+                # First non-ghost after a chain of ghosts terminates the
+                # virtual extension at its ``start_release_id``.
+                return sibling.start_release_id
+            # A non-ghost adjoins ``mv``'s end directly (no ghost bridge),
+            # so no extension applies — keep the schema value.
+            return mv.end_release_id
+        if saw_ghost:
+            return None
+        return mv.end_release_id
+
```
ruizmaa, line 605: `saw_ghost=True` here doesn't mean "unbounded", it means the last ghost has a determinate `end_release_id` that's tracked in `boundary` and then discarded, returning `None` instead. That silently disables the "past the end" check for any release beyond it.

A ghost with a determinate end and no sibling after it lets `_resolve_release("MODG", "1.0.0", "4.3")` succeed when it should raise:

```python
from datetime import date

from dpmcore.orm import Base
from dpmcore.orm.packaging import ModuleVersion

session.add_all(
    [
        Release(release_id=3, code="4.0", date=date(2025, 1, 1)),
        Release(release_id=1010000003, code="4.2", date=date(2026, 3, 1)),
        Release(release_id=5, code="4.2.1", date=date(2026, 6, 1)),
        Release(release_id=7, code="4.3", date=date(2026, 9, 1)),
        # MODG 1.0.0 (non-ghost) covers [4.0, 4.2)
        ModuleVersion(
            module_vid=200, module_id=99, code="MODG", version_number="1.0.0",
            start_release_id=3, end_release_id=1010000003,
            from_reference_date=date(2025, 1, 31), to_reference_date=date(2026, 3, 30),
        ),
        # MODG 1.1.0 (ghost) covers [4.2, 4.2.1) -- and has no sibling after it
        ModuleVersion(
            module_vid=201, module_id=99, code="MODG", version_number="1.1.0",
            start_release_id=1010000003, end_release_id=5,
            from_reference_date=date(2026, 3, 31), to_reference_date=date(2026, 3, 31),
        ),
    ]
)
session.flush()

# Neither version covers release "4.3" -> this should raise ValueError
# ("... is past the end of module version ..."). It doesn't.
mv, rel = svc._resolve_release("MODG", "1.0.0", "4.3")
print(f"BUG: accepted release {rel.code!r} against {mv.version_number} with no error")

```

ruizmaa (approved):

src/dpmcore/services/ast_generator.py
victorprietosandamaso-cloud: You're right — inlined it in _siblings_past_end in commit e22b10d. A three-condition check used in one place didn't earn its own function.

src/dpmcore/services/ast_generator.py
victorprietosandamaso-cloud: Confirmed the repro (MODG with ghost [4.2, 4.2.1) and no sibling after it — release 4.3 was silently accepted). Fixed in e22b10d:
- Track last_ghost_end_id alongside boundary inside the walk, updated only when a bounded ghost genuinely extends the boundary.
- When the loop exits with saw_ghost=True and no non-ghost was hit, return last_ghost_end_id instead of None. An open-ended ghost still returns None from inside the loop (sibling.end_release_id is None), so that path is preserved.
Added test_extension_stops_at_bounded_ghost_end_when_no_sibling covering the exact reproduction from your comment (MODG 1.0.0 [4.0, 4.2), ghost 1.1.0 [4.2, 4.2.1), no sibling after) —  verifies 4.2 passes, 4.2.1 and 4.3 both raise "past the end". Full suite (19 tests) green

PR #235, opened 2026-07-22
feat(dpm-xl): support setof(<recordset>, <property>) selection

### LINKED ISSUE #234: feat(dpm-xl): support setof(<recordset>, <property>) selection operator
### Initial Checks

### Problem Description

DPM-XL should provide a setof(<recordset>, <property>) operator that projects a recordset onto a single component (property/dimension) and returns the corresponding scalar set. It is the natural companion of in / setdiff when the RHS
  needs to be extracted from another table rather than enumerated by hand.

  The token SET_OF already exists in src/dpmcore/dpm_xl/grammar/dpm_xlLexer.g4, but there is no matching parser production, no AST node, and no visitor, so any expression that uses it aborts at parse time. Published EGDQ validation
  formulas use this operator in exactly this shape:

  not({tC_28.00, c020}[where f = [eba_ZZ:x1]][get INC] in setof({tC_29.00}, INC))

  Semantics: setof({tC_29.00}, INC) returns the scalar set of all values of the INC property present in table C_29.00, so the outer in check verifies that every reported INC value in C_28.00 (filtered by fact [where f = ...]) also exists
  in C_29.00.

  Affected EGDQ checks (6 unique VR codes): EGDQ_0070, EGDQ_0071a, EGDQ_0071b, EGDQ_0131, EGDQ_0132, EGDQ_0133.

  Reproduction of the current failure:

  ```python

  r = SyntaxService().validate('setof({tC_29.00}, INC)')
  # r.is_valid is False
  # r.error_message: "extraneous input 'setof' expecting
  #                   {BOOLEAN_LITERAL, NOT, MATCH, 'with', ...}"
  ```

### Proposed Solution

```python
Wire the missing parser rule, AST node, and semantic pass:

  1. Add a production to the parser under setExpression, aligned with the existing shape of union, intersect, setdiff, symdiff:

setExpression:

SETDIFF LPAREN left=expression COMMA right=expression RPAREN      #setdiffSetExpr
SYMDIFF LPAREN left=expression COMMA right=expression RPAREN      #symdiffSetExpr
SET_OF LPAREN recordset=expression COMMA property=keyNames RPAREN #setOfExpr
      
  2. Regenerate the ANTLR parser (antlr4 -Dlanguage=Python3 -visitor per the project's CLAUDE.md pin at 4.9.2).
  3. Add a SetOfOp AST node in dpm_xl/ast/nodes.py with attributes operand (recordset expression) and component (property code). Follow the existing shape of the GetOp node — it already stores an operand + a component and is emitted by
  [get X].
  4. Add visitSetOfOp in dpm_xl/ast/constructor.py and visit_SetOfOp in dpm_xl/semantic_analyzer.py. Semantics: validate that the recordset resolves against a real table version, validate that property is a component defined on that
  table's key axis or resolves as an implicit / GetOp component through ViewOpenKeysQuery, and type the result as a Set whose element type matches the property's DataType.
  5. Wire the ML/engine step in dpm_xl/ast/ml_generation.py::visit_SetOfOp — persist an OperationNode with the correct operator id (a new row in the Operator catalogue via a Django/ORM migration; the recent set-operator alignment PR #203
  already added rows for the sibling set operators, so this can piggyback on the same shape).
  6. Reject invalid shapes with a proper SemanticError (not a bare Exception) — mirror the guard added in feat(dpm_xl): first-class set operators #193.

  Expected AST for setof({tC_29.00}, INC):

  Start
  └── SetOfOp(component='INC')
      └── VarID(table='C_29.00', rows=None, cols=None, sheets=None)

  Test coverage to add:
  - Unit: SyntaxService().parse('setof({tX}, PROP)') yields the expected AST shape.
  - Unit: semantic validation over a memory-session fixture where PROP is a resolvable key/property.
  - Integration: end-to-end via SemanticService.validate and ScopeCalculatorService.calculate_from_expression for the EGDQ_0070 reproducer above.
  - Reject: setof({tX}) (missing property), setof({tX}, ) (empty property), setof(1, PROP) (non-recordset operand) — all raise SemanticError with a typed code.
```

### Alternatives Considered

 - Expressing the same intent by materialising the set inline ({r1, r2, r3}) — rejected, because the affected checks target dynamic sets extracted at runtime from another table, not enumerated literals.
  - Overloading [get PROP] to also flatten to a set — rejected, because [get] operates on a single row/cell and returning a scalar set would break the type contract downstream.

### Additional Context

The SET_OF token was introduced alongside the other first-class set operators (#193 / cr-192, #203 / cr-202) but the parser rule, AST node and visitor were never landed.

victorprietosandamaso-cloud:
## Summary
  - Extends the existing `set_of` unary operator to accept a two-argument form `set_of(recordset, component)` that projects the recordset onto a single component (property/dimension) and returns its scalar set.
  - Accepts `setof` (no underscore) as an alias of `set_of` — both spellings map to the same `SET_OF` token, so historical or shorthand syntax parses interchangeably.
  - Grammar regenerated with ANTLR 4.9.2 per the `CLAUDE.md` pin. Both the parser and lexer generated files are refreshed.
  - AST `SetOfOp` gains an optional `component: str | None` field; the pre-existing unary shape (`component=None`) is unchanged.
  - `visitSetOfProjectExpr` in the constructor builds the two-argument node and strips backticks from an escaped-identifier component (e.g. `` `f` `` for the fact component).
  - `visit_SetOfOp` in the semantic analyser resolves the component against the operand's `Structure` and types the resulting `ScalarSet` by that component's scalar type; a missing component raises the standard `1-5` (open-keys not found)
  error.
  - `visit_SetOfOp` in `ml_generation` emits an extra `component` operand reference (property id resolved through `ItemCategoryQuery`) whenever the two-argument form is used, mirroring how `visit_GetOp` serialises `[get X]`.
  - Enables the `EGDQ_0070`, `EGDQ_0071a`, `EGDQ_0071b`, `EGDQ_0131`, `EGDQ_0132`, `EGDQ_0133` family of checks to parse and build a valid AST.

  ## Test plan
  - [x] `poetry run pytest tests/unit/ast/test_set_operators.py` — 46/46 green, including 5 new tests: `set_of` unary keeps working, `setof` alias parses identically, both spellings accept the two-argument form, backticks are stripped
  from the component, and JSON round-trip carries the new field.
  - [x] `poetry run pytest tests/unit --ignore=tests/unit/cli` — 1461/1461 green, no regressions.
  - [x] `poetry run ruff format` + `ruff check` clean on every touched file.
  - [x] `poetry run mypy` clean on the four source modules touched (`ast/nodes.py`, `ast/constructor.py`, `semantic_analyzer.py`, `ast/ml_generation.py`).
  - [x] End-to-end parse smoke: the `EGDQ_0070` reproducer `not({tC_28.00, c020}[where f = [eba_ZZ:x1]][get INC] in setof({tC_29.00}, INC))` yields `SetOfOp(component='INC', operand=VarID(table='C_29.00'))` without any parser error.

  Fixes #234


src/dpmcore/dpm_xl/grammar/dpm_xlLexer.g4
```diff
@@ -422,7 +422,7 @@ CLAUSE_DAY:                    'day' -> type(DAY);
 CLAUSE_DATE:                   'date' -> type(DATE);
 
 // Set operators
-CLAUSE_SET_OF:                 'set_of'    -> type(SET_OF);
+CLAUSE_SET_OF:                 ('set_of' | 'setof')    -> type(SET_OF);
```
ruizmaa, line 425: `dpm-xl-docs` only has `set_of`, but not `setof`. So is this okay? Same question in line 74.

victorprietosandamaso-cloud:
Closing without merging. The two-argument setof(recordset, property) form is already expressible with the existing [get <property>] operator, so no grammar change is needed. The setof alias (no underscore) doesn't match dpm-xl-docs and is not being added.

PR #253, opened 2026-07-27
fix(scripts): declare cross-validation operands in dependency_modules

### LINKED ISSUE #250: Generated dependency_modules over-declares entire dependency module instead of only referenced tables/variables
## Summary

When a generated module script declares a **cross-instance dependency**, the `dependency_modules` section contains the **entire dependency module** (all tables + all variables) instead of only the tables/variables actually referenced by the cross-rules. This over-declaration corrupts the adam-engine's combined cross-module data frame at execution time: the **primary (home) module's fact values are dropped**, so home operands resolve to `null`.

The visible symptoms in the engine are:
- bare single-cell home operands → `Error in root_artifacts: Artifact { reason: "Scalar can't be created for this data" }`
- aggregated home operands → silently computed as `0` (wrong result, no error)

## How native/correct scripts declare it

Native EBA scripts declare in `dependency_modules[<dep>]` only the handful of tables/variables their cross-rules use. Example — `validation-scripts/EBA/GSII-1.1.0.json`:

```
dependency_modules[finrep9] = { tables: {F_22.02, F_14.00}, variables: {…11 ids…} }   # 2 tables, 11 variables
```

## The defective output

A generated `COREP_OF` script that has cross-rules depending on `finrep9` instead emits:

```
dependency_modules[finrep9] = 122 tables, 13,394 variables   # the ENTIRE finrep9 module
dependency_modules[ae]      = 23 tables, 1,226 variables      # the ENTIRE ae module
```

The declared content is *internally correct* (open_keys, variable types all match the real finrep9 module) — the problem is purely that **the whole module is declared rather than the referenced subset**.

## Minimal reproduction (proves it is the declaration)

Engine: `adam-engine 1.2.0-pre8`.

1. Run native `validation-scripts/EBA/GSII-1.1.0.json` on the GSII example instance with the finrep9 example instance as dependency, validating `v10862_m` (`{tG_01.00} = {tF_22.02}`).
   → **`warning`** (executes correctly; home `G_01.00` value present).

2. Replace *only* `dependency_modules[finrep9]` in that same GSII script with the full 122-table / 13,394-variable finrep9 declaration (everything else identical), and re-run.
   → **`exception`: "Scalar can't be created for this data"** — the home operand data is dropped.

The single change (subset declaration → full-module declaration) flips a correct run into the failure.

## Expected behavior

When generating a module script, `dependency_modules[<dep>]` should contain **only the tables and variables (datapoints) referenced by the operations that depend on that module** — matching the native EBA scripts — not the entire dependency module.

## Notes

- This is the dependency-declaration/generation half of the problem. Separately, the adam-engine also appears to drop home data for a `COREP` (home, `C_` tables) + `finrep9`/`ae` (dependency, `F_` tables) pairing even after the declaration is reduced to a single table — that part looks like an engine-side issue and is being reported separately.

### LINKED ISSUE #251: dependency_modules[<dep>].variables omits home-module variables referenced by cross-validations, breaking cross-module execution
## Summary

When generating a module script that has **cross-instance validations**, the `dependency_modules[<dep>].variables` map must contain **every variable (datapoint) referenced by those cross-validations — including the ones that belong to the _home_ (primary) module**, not only the ones belonging to the dependency module.

Currently the generator populates `dependency_modules[<dep>].variables` with **only the dependency module's variables** and omits the home-module operand variables. The adam-engine uses this field to resolve/assign operands in the combined cross-module frame, so with the home variables missing it cannot build the home operands:

- bare single-cell home operand → `Error in root_artifacts: Artifact { reason: "Scalar can't be created for this data" }`
- home operand inside an aggregation → silently computed as `0` (wrong result, no error)

This was confirmed by the engine team as the root cause.

## Correct vs. defective output

Rule (COREP home, finrep9 dependency):
```
PwC_602_1:  {tC_01.00, r0040, c0010, default:0} <= {tF_01.03, r0020, c0010, default:0}
            └── home operand (var 32673) ──┘        └── dependency operand (var 56987) ──┘
```

**Correct** — native EBA scripts include the home variable in the dependency's `variables` map:
- `GSII-1.1.0.json` → `dependency_modules[finrep9].variables` = 11 ids = 9 finrep9 vars **+ 2 home GSII vars** (459089, 456246).
- `COREP_OF-3.1.0.json` → `dependency_modules[corep_le].variables` includes the home `C_01.00` vars used by v0655_m.

**Defective** — the generated script omits them:
- `dependency_modules[finrep9].variables` = 13,394 ids, **all finrep9**, home var `32673` **absent**.
- `dependency_modules[ae].variables` = 1,226 ids, all ae, home vars absent.

## Fix (what the generator must do)

For each dependency module `D` referenced by `cross_instance_dependencies`:

1. Gather every operand datapoint from the operations affected by `D` (i.e. walk the `VarID` nodes of each cross-rule and collect `data[].datapoint`).
2. Ensure **all** of those datapoints appear as keys in `dependency_modules[D].variables`, with their data-type value (`m`/`e`/`i`/…):
   - datapoints whose table belongs to `D` → type from `D`'s module definition (already handled today);
   - datapoints whose table belongs to the **home module** → type looked up from the **home module's** variable definitions (this is the part currently missing).

`dependency_modules[D].tables` stays as-is (it lists the dependency tables). Only the module-level `variables` map needs the home variables added.

Pseudocode:
```
for dep in cross_instance_dependencies:
    for op in dep.affected_operations:
        for datapoint in operand_datapoints(op):        # all VarID data[].datapoint of the rule
            if datapoint not in dependency_modules[dep].variables:
                dependency_modules[dep].variables[datapoint] = variable_type(datapoint)   # from home or dep module
```

## Verification

Post-condition to assert in the generator/tests: **for every cross-instance validation, all of its operand datapoints are present in `dependency_modules[<its dependency>].variables`.**

Empirical confirmation (adam-engine 1.2.0-pre8, real COREP_OF + finrep9 run, 218 operations):

| | exceptions | of which `"Scalar can't be created"` | Ok |
|---|---|---|---|
| generated script (home vars missing) | 49 | **43** | 163 |
| after adding the 50 finrep9 + 192 ae home vars | 6 | **0** | 199 |

All 43 `"Scalar can't be created"` exceptions cleared and home operands resolved to their real values (e.g. a `sum({C_04.00,…})` that had wrongly computed `0` now returns the reported ≈10⁷ figure).

## Relationship to #250

#250 is about the *dependency-side* over-declaration (emitting the entire dependency module instead of the used subset). This issue is the complementary, higher-priority half: whatever the dependency side contains, the `variables` map must **also include the home-module variables** used by the cross-rules — that omission is what actually breaks execution. The engine team has indicated they prefer to eventually move dependency-variable filtering into the engine; until then, generated scripts must carry the full set of cross-validation variables (home + dependency) in this field.

antonio-olleros:
The adam-engine resolves every operand of a cross-instance validation against dependency_modules[<dep>].variables, home-module operands included. The generator declared only the dependency module's own variables, so a home operand could not be built: a bare single cell failed with "Scalar can't be created for this data" and an aggregated one silently computed 0 (#251).

Collect every operand datapoint (and its type) per operation from the raw serialised AST — before _clean_ast_data_entries strips data_type — and graft them into each dependency module's variables map. The dependency module's own definition of a datapoint still wins.

While declaring the referenced subset, drop the over-declaration: a dependency module was emitted whole (122 tables / 13k variables where native EBA scripts declare a handful), so restrict its tables to the ones the cross-rules reference and each table's variables to the referenced datapoints, keeping open_keys intact (#250). If narrowing would leave nothing declarable the module is kept whole rather than dropped — over-declaring is wrong, losing a genuine cross-instance dependency is worse.

_merge_dep_modules now unions a repeated table's variables instead of keeping the first entry: with narrowing, two operations touching different cells of the same dependency table each declare only their own datapoints and both must survive the merge.

Closes #251
Closes #250

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/services/scope_calculator.py
+++ b/src/dpmcore/services/scope_calculator.py
@@ -411,6 +411,17 @@ class ScopeCalculatorService:
         if not tables_dict:
             return None
 
+        # The timeshift is a module-level property carried by the dependency
+        # module's tables. Compute it BEFORE narrowing: narrowing drops any
+        # table not referenced by the cross-rules, and a dropped table takes
+        # its timeshift with it — a module whose only timeshifted table is
+        # not referenced would otherwise fall back to ref_period T.
+        ref_period = "T"
+        for tbl_code in tables_dict:
+            rp = ts.get(tbl_code)
+            if rp and rp != "T":
+                ref_period = rp
+
         # #250: declare only the tables and datapoints the cross-rules
         # actually reference — a whole dependency module is 100+ tables and
         # 10k+ variables, where native EBA scripts declare a handful.
@@ -419,12 +430,6 @@ class ScopeCalculatorService:
         )
         if narrowed:
             tables_dict = narrowed
-
-        ref_period = "T"
-        for tbl_code in tables_dict:
-            rp = ts.get(tbl_code)
-            if rp and rp != "T":
-                ref_period = rp
 
         module_entry: Dict[str, Any] = {
             "URI": uri,
--- a/src/dpmcore/services/scope_calculator.py
+++ b/src/dpmcore/services/scope_calculator.py
@@ -237,6 +237,8 @@ class ScopeCalculatorService:
         time_shifts: Optional[Dict[str, str]] = None,
         compute_alternative_deps: bool = True,
         release_code: Optional[str] = None,
+        referenced_variables: Optional[Dict[str, str]] = None,
+        referenced_tables: Optional[Set[str]] = None,
     ) -> Dict[str, Any]:
         """Build dependency information for a scope result.
 
@@ -256,6 +258,15 @@ class ScopeCalculatorService:
             release_code: Optional release code; resolved to
                 ``release_id`` via :class:`Release.code`. Mutually
                 exclusive with ``release_id``.
+            referenced_variables: Optional ``{datapoint: type_code}`` of
+                every operand datapoint the operation references, across
+                all modules it spans — home module included. Declared in
+                each dependency module's ``variables`` map (#251).
+            referenced_tables: Optional table codes the operation
+                references. Together with ``referenced_variables`` this
+                narrows each dependency module's declaration to the
+                subset the operation uses (#250); omit both to declare
+                the dependency modules whole.
 
         Returns a dict with:
         - ``intra_instance_validations``
@@ -343,6 +354,8 @@ class ScopeCalculatorService:
                 release_id=release_id,
                 ts=ts,
                 operation_code=operation_code,
+                referenced_variables=referenced_variables,
+                referenced_tables=referenced_tables,
             )
             if entry is None:
                 continue
@@ -375,6 +388,8 @@ class ScopeCalculatorService:
         release_id: Optional[int],
         ts: Dict[str, str],
         operation_code: Optional[str],
+        referenced_variables: Optional[Dict[str, str]] = None,
+        referenced_tables: Optional[Set[str]] = None,
     ) -> Optional[Tuple[Dict[str, Any], str, Dict[str, Any]]]:
         """Build a single (cross_dep, uri, dependency_module) triple.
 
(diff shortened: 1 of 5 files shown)


src/dpmcore/services/scope_calculator.py
```diff
@@ -396,6 +411,15 @@ def _build_dependency_entry(
         if not tables_dict:
             return None
 
+        # #250: declare only the tables and datapoints the cross-rules
+        # actually reference — a whole dependency module is 100+ tables and
+        # 10k+ variables, where native EBA scripts declare a handful.
+        narrowed = self._narrow_dependency_tables(
```
andres-sole, line 428: The narrowing happens before the ts.get(tbl_code) loop, so a table dropped by narrowing takes its timeshift with it and the whole dependency module falls back to ref_period T

andres-sole (approved):

PR #255, opened 2026-07-27
fix(dpm-xl): tighten default-value semantic checks around isnull

### LINKED ISSUE #254: Semantic analyzer accepts isnull(x) when x has a non-null default, and accepts numeric default: on String cells
### Initial Checks

### Summary

Two independent gaps in the semantic analyzer allow a nonsensical expression to pass validation:

  not(isnull({tC_34.06, c0010, default:0}))
where c0010 is declared String.

  1. default: type check is too permissive on String cells. SemanticAnalyzer.__check_default_value (src/dpmcore/dpm_xl/semantic_analyzer.py:249-272) checks the default against the cell type using implicit_type_promotion_dict (src/dpmcore/dpm_xl/types/promotion.py:23-32), which promotes every scalar to String. As a result, default:0 (Integer) on a String cell is accepted. Thecheck only fires in the reverse direction (e.g. default:"foo" on Number).
  2. isnull(x) does not warn when x has a non-null default. IsNull (src/dpmcore/dpm_xl/operators/comparison.py:21-25) inherits _BaseUnary with no operand inspection. When the operand carries
  a non-null default:, the call is tautologically false (and not(isnull(...)) tautologically true), but no warning or error is emitted.

### Reproducible Example

```python
from dpmcore import connect
  # validate: not(isnull({tC_34.06, c0010, default:0}))
  # c0010 declared as String
  # Expected: SemanticError (numeric default on String) and/or warning (isnull with non-null default is always false)
  # Actual:   passes semantic analysis silently
No existing test covers isnull({..., default:<non-null>}); tests/unit/ast/test_default_null.py only covers default:null.
```

### Expected vs Actual

Expected:
  - SemanticError on default:0 when the cell is String (type mismatch).
  - Warning (or error) on isnull(x) when x has a non-null default:, since the result is always false.

  Actual:
  - Expression passes semantic analysis silently.

### dpmcore version

0.1.2rc3

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

victorprietosandamaso-cloud:
Two independent gaps in the semantic analyzer let a nonsensical
  expression like `not(isnull({tC_34.06, c0010, default:0}))` — with
  c0010 a String cell — pass validation silently.

  1. `__check_default_value` validated defaults against the cell type via
     `implicit_type_promotion_dict`, whose "everything → String" fallback
     accepted `default:0` on a String cell. Guard the String cell case
     explicitly: only String and Null defaults are meaningful there.

  2. `IsNull` inherited `_BaseUnary` with no operand inspection, so
     `isnull({..., default:<non-null>})` — tautologically false — passed
     without any warning. `visit_UnaryOp` now inspects the operand's AST
     default before visiting and emits a semantic warning.

  Closes #254

  ## Summary

  - Reject numeric/boolean/item defaults on String cells (SemanticError 3-6).
  - Warn on `isnull(x)` when `x`'s default is non-null.

  ## Checklist

  - [x] Tests pass (`pytest tests/unit/semantic tests/unit/dpm_xl tests/unit/ast`)

  ## Impact / Risk

  - Breaking-ish: expressions currently passing with `default:<numeric>`
    on a String cell now raise SemanticError 3-6. This is the intended
    behaviour per issue #254.
  - No schema/migration concerns.

### DIFF
--- a/src/dpmcore/dpm_xl/semantic_analyzer.py
+++ b/src/dpmcore/dpm_xl/semantic_analyzer.py
@@ -211,21 +211,8 @@ class InputAnalyzer(ASTTemplate, ABC):
         self, node: UnaryOp
     ) -> Operand:
         op = cast(str, node.op)
-        # ``isnull(x)`` on a selection with a non-null default is tautologically
-        # false (the default guarantees the operand is never null). Warn so the
-        # user notices the modelling mistake instead of relying on ``not(...)``
-        # around a constant. ``default:null`` is materialised as a ``Constant``
-        # with ``type_ == "Null"`` (see visitDefault in the AST constructor).
         if op == ISNULL:
-            default = getattr(node.operand, "default", None)
-            if (
-                default is not None
-                and getattr(default, "type", None) != "Null"
-            ):
-                add_semantic_warning(
-                    "isnull(...) on a selection with a non-null default is "
-                    "always false"
-                )
+            self.__warn_if_isnull_has_non_null_default(node)
         operand_symbol = self.visit(node.operand)
         if op in UNARY_OP_MAPPING:
             result = UNARY_OP_MAPPING[op].validate_types(operand_symbol)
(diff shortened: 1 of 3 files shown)


src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -208,8 +210,23 @@ def visit_BinOp(  # type: ignore[override]
...
         self, node: UnaryOp
     ) -> Operand:
-        operand_symbol = self.visit(node.operand)
         op = cast(str, node.op)
+        # ``isnull(x)`` on a selection with a non-null default is tautologically
+        # false (the default guarantees the operand is never null). Warn so the
+        # user notices the modelling mistake instead of relying on ``not(...)``
+        # around a constant. ``default:null`` is materialised as a ``Constant``
+        # with ``type_ == "Null"`` (see visitDefault in the AST constructor).
+        if op == ISNULL:
```
andres-sole, line 214: The check only looks at the immediate operand, one pair of parentheses defeats it: `isnull( ({c0040, default: 0}) )` produces no warning

src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -208,8 +210,23 @@ def visit_BinOp(  # type: ignore[override]
...
+            default = getattr(node.operand, "default", None)
+            if (
+                default is not None
+                and getattr(default, "type", None) != "Null"
+            ):
+                add_semantic_warning(
```
andres-sole: The message says nothing about which selection is at fault, and it repeats per occurrence. `with {tC_105.01, default: 0, interval: true}: isnull({c0040}) and isnull({c0130})` currently yields two identical lines

andres-sole (approved):

PR #264, opened 2026-07-29
fix: fixed ecb ids drift between eba dpm releases

### LINKED ISSUE #263: ECB validation IDs drift between EBA/DPM releases
## Initial Checks

## Summary

ECB-owned rows in `Operation`, `OperationVersion`, `OperationNode`, and
`OperandReference` should keep stable IDs across EBA/DPM releases, since
the ECB validation set itself does not change between releases — only
the underlying EBA/DPM dataset does. Instead, these IDs shift forward on
every release, in lockstep with the growth of the unrelated EBA dataset.

## Evidence

Compared the ECB portion of three consecutive DPM databases
(`dpm-database-4.2.1.sqlite`, `dpm-database-4.3.sqlite`, `dpm-database-4.4.sqlite`,
generated from the corresponding `.accdb` sources). The ECB `Organisation`
row is stable (`OrgID = 1024` in all three) and the ECB operation count
is identical (1164 rows) in all three, but the assigned IDs are not:

| Code | 4.2.1 | 4.3 | 4.4 |
|---|---|---|---|
| `Operation.OperationID` (EGDQ_0022_1) | 1010000016 | 1010000719 | 1010000900 |
| `OperationVersion.OperationVID` (EGDQ_0022_1) | 1010000035 | 1010000728 | 1010000937 |

Root cause: the max `OperationID` among *non*-ECB (i.e. native EBA)
operations grows every release —

- 4.2.1: max = 1,010,000,013
- 4.3: max = 1,010,000,716 (+703)
- 4.4: max = 1,010,000,897 (+181)

— and the ECB import assigns IDs as `MAX(id) + 1` over the whole table,
so the ECB block always starts right after wherever the EBA dataset
currently ends. Any new EBA rows added ahead of it push the entire ECB
block forward.

The same pattern shows up in `OperationNode.NodeID` (a table shared by
the whole engine, not just ECB), with even larger jumps:

- 4.2.1: ECB range 1,010,000,166–1,010,001,251
- 4.3: ECB range 1,010,002,378–1,010,003,463
- 4.4: ECB range 1,010,004,168–1,010,005,253

Image

Image

Image

## Impact

Any external reference to these IDs (e.g. saved reports, cross-links,
diffing tools, downstream consumers pinning to specific operation IDs)
breaks silently between releases even though nothing about the ECB
validation set changed.

## Reproducible Example

```python

from dpmcore.orm.base import Base
from dpmcore.orm.operations import Operation
from dpmcore.services.ecb_validations_import import EcbValidationsImportService

ECB_CSV = "ecb.csv"  # contents: "vr_code,start_release\nV1,4.0\n"


def assigned_operation_id(eba_operation_count: int) -> int:

    # Simulate the EBA/DPM operations that already exist before ECB import runs
    for i in range(1, eba_operation_count + 1):
        session.add(
            Operation(
                operation_id=i,
                code=f"EBA_{i}",
                type="validation",
                source="hierarchy",
            )
        )
    session.commit()
    session.close()

    EcbValidationsImportService(engine).import_csv(ECB_CSV)

    op = session.query(Operation).filter(Operation.code == "V1").one()
    session.close()
    return op.operation_id


# "release N": 10 EBA operations exist before ECB import runs
release_n_id = assigned_operation_id(10)

# "release N+1": EBA dataset grew to 5000 operations, but the ECB CSV is identical
release_n_plus_1_id = assigned_operation_id(5000)

print(release_n_id, release_n_plus_1_id)
```

## Expected vs Actual

Expected: `release_n_id == release_n_plus_1_id` — the ECB validation "V1" is
unchanged between releases, so its `OperationID` (and `OperationVID`, `NodeID`,
`OperandReferenceID`) should stay the same regardless of how many unrelated
EBA operations were added ahead of it.

Actual: `release_n_plus_1_id - release_n_id == 4990`. `EcbValidationsImportService`
computes each ID as `MAX(id) + 1` over the *whole* `Operation`/`OperationVersion`/
`OperationNode`/`OperandReference` tables instead of anchoring to a reserved ECB
block, so the entire ECB ID range drifts forward by however much the EBA
dataset grew since the last release.

Confirmed against real DPM databases too: `EGDQ_0022_1`'s `OperationID` was
`1010000016` in DPM 4.2.1, `1010000719` in 4.3, and `1010000900` in 4.4 —
despite the ECB validation rule itself being unchanged.

## dpmcore version

0.1.2rc5

## Python version

3.12.3

## Database backend

SQLite

## Additional Context

A local fix anchoring ECB-assigned IDs to a fixed floor
(`EcbValidationsImportService._ECB_ID_BASE`) above the current EBA/DPM
dataset size is in progress in `src/dpmcore/services/ecb_validations_import.py`.

guillermo-garcia-1:
## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/services/ecb_validations_import.py
+++ b/src/dpmcore/services/ecb_validations_import.py
@@ -205,8 +205,13 @@ class EcbValidationsImportService:
         Without ``floor`` this is a plain "max + 1" over the whole table,
         which drifts release to release as the surrounding EBA/DPM dataset
         grows. Passing ``floor`` scopes the scan to IDs already inside that
-        reserved block, so the first assigned ID stays the same run after
-        run regardless of how large the rest of the table has become.
+        reserved block, so the first ID handed out from it stays the same
+        regardless of how large the rest of the table has become.
+
+        This does not guarantee a stable ID per ECB code across a full
+        reimport: codes are enumerated via ``sorted(unique_codes)``, so
+        adding a new code that sorts before existing ones shifts every
+        code after it.
         """
         column = getattr(model, attr_name)
         query = session.query(func.max(column))
--- a/src/dpmcore/services/ecb_validations_import.py
+++ b/src/dpmcore/services/ecb_validations_import.py
@@ -10,7 +10,7 @@ from hashlib import md5
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Set
 
-from sqlalchemy import func
+from sqlalchemy import func, or_
 from sqlalchemy.engine import Engine
 from sqlalchemy.orm import Session, sessionmaker
 
@@ -72,9 +72,11 @@ class EcbValidationsImportResult:
 class EcbValidationsImportService:
     """Import ECB validation rules from a CSV export into a DPM database."""
 
-    # Fixed floor for every ID this service assigns, so ECB's IDs stay put
-    # across releases instead of drifting with the native EBA/DPM dataset.
-    _ECB_ID_BASE = 1_030_000_000
+    # Pinned instead of computed via max+1, so they no longer drift as
+    # organisations/prefixes are added upstream.
+    _ECB_ID_PREFIX = 103
+    _ECB_ID_BASE = _ECB_ID_PREFIX * 10_000_000
+    _ECB_ORG_ID = 1_030_000
 
     # Required headroom between the native max ID and `_ECB_ID_BASE`.
     _ECB_ID_BASE_MIN_MARGIN = 1_000_000
@@ -274,14 +276,27 @@ class EcbValidationsImportService:
         if org is not None:
             return org
 
-        next_org_id = self._next_int_id(session, Organisation, "org_id")
-        max_prefix = session.query(func.max(Organisation.id_prefix)).scalar()
+        conflicting = (
+            session.query(Organisation)
+            .filter(
+                or_(
+                    Organisation.org_id == self._ECB_ORG_ID,
+                    Organisation.id_prefix == self._ECB_ID_PREFIX,
+                )
+            )
+            .first()
+        )
+        if conflicting is not None:
+            raise EcbValidationsImportError(
+                f"Organisation {conflicting.acronym!r} already occupies"
+                " the reserved ECB org_id/id_prefix."
+            )
 
         org = Organisation(
-            org_id=next_org_id,
+            org_id=self._ECB_ORG_ID,
             name="European Central Bank",
             acronym="ECB",
-            id_prefix=int(max_prefix or 0) + 1,
+            id_prefix=self._ECB_ID_PREFIX,
         )
         session.add(org)
         session.flush()
--- a/src/dpmcore/services/ecb_validations_import.py
+++ b/src/dpmcore/services/ecb_validations_import.py
@@ -72,6 +72,13 @@ class EcbValidationsImportResult:
 class EcbValidationsImportService:
     """Import ECB validation rules from a CSV export into a DPM database."""
 
+    # Fixed floor for every ID this service assigns, so ECB's IDs stay put
+    # across releases instead of drifting with the native EBA/DPM dataset.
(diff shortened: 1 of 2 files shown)


andres-sole (commented):
ECB org_id is still assigned with plain max+1, so it still drifts. The real 4.4 database has 3 organisations (1001, 1012, 1023), so ECB gets 1024, and it gets a different value as soon as an organisation is added upstream. That value is written to Operation.owner_id, and it is also part of the input to the concept GUID, so Concept.concept_guid and Operation.row_guid change between releases too.

To fix pin the ECB org_id, or remove org_id from the GUID input and use the fixed "ECB" acronym.

The base is also hardcoded while the id_prefix next to it is computed. DPM allocates IDs as IDPrefix * 10000000 + sequence: EBA has prefix 101 and its Operation IDs run from 1010000004 to 1010000897, and Release IDs follow the same rule (4.2.1 is 1010000003). The existing prefixes are 100, 101 and 102, so ECB gets 103, and 103 * 10000000 is exactly _ECB_ID_BASE. That only matches because of the max+1. If an organisation is added upstream and takes 103, ECB moves to 104 while the IDs still come from block 103, and that organisation writes rows inside the ECB block.

To fix pin the ECB id_prefix as well and derive the base from it, id_prefix * 10000000, so the prefix and the block cannot disagree.

Also the margin check only reads IDs below the floor, so it cannot detect native IDs above the floor. Test with a native operation_id of 1040000000, the check passes and the ECB operation gets 1040000001. The drift returns and nothing is reported. With the prefix problem above this case is realistic, not only theoretical. To fix fail or warn when rows at or above the floor were not written by ECB. For Operation, compare owner_id with the ECB org id. The other three tables need a join through operation_id.

The block itself is free today. I checked the 4.4 database and there are no rows at or above 1030000000 in Operation, OperationVersion, OperationNode or OperandReference, with about 20 million left to the native maximum.

OperationScope.operation_scope_id is not assigned here, and not in the scope calculator either, so the database assigns it as max+1 over the whole table and those IDs still drift. row_guid is stable because it comes from the operation_vid this PR fixes, so it may not matter. Is that intended, or should scopes go in the block too?

For follow up:
- ecb_validations_import.py:230 - "only ECB writes IDs >= floor" is the exact assumption that breaks if prefix 103 is taken upstream
- ecb_validations_import.py:241 - message hardcodes _ECB_ID_BASE but the function takes floor; tests pass 1000 and match on "_ECB_ID_BASE". Also: bumping the base is a planned drift, and Integer caps at 2147483647
- ecb_validations_import.py:205 - docstring overstates the guarantee; per-code IDs still shift with sorted(unique_codes) at :642
- ecb_validations_import.py:77 - databases imported with the old code end up with ECB IDs in two ranges; the PR description should state whether they need a fresh import. The reserved range should be documented in the module docstring
- ecb_validations_import.py:55 - error derives from Exception, not DpmCoreError (pre-existing)
- test file:649 - also assert the value equals _ECB_ID_BASE
- test file:314 - check is only tested through Organisation, never through the four real models
- test file:621 - 5000 rows inserted one at a time; that scale is not needed
- test file:630 - session not wrapped in try/finally, unlike the rest of the file

guillermo-garcia-1:
## Response to review

- Pinned ECB `org_id`/`id_prefix` instead of `max+1`; `_ECB_ID_BASE` now derives from the prefix. Fixes the GUID drift too.
- Margin check is now bidirectional: also fails if a non-ECB row already sits at/above the floor.
- `OperationScope.operation_scope_id` now comes from the same reserved block as the other tables.
- Docstrings updated: reserved block, two-range caveat, and the per-code `sorted(unique_codes)` limitation.
- `EcbValidationsImportError` now derives from `DpmCoreError`.
- Tests reworked to use the real models instead of `Organisation`, plus coverage for everything above.
- Verified against real 4.2.1/4.3/4.4 databases: only the touched tables differ, row-for-row, and ECB IDs are now identical across all three releases.

andres-sole (approved):

PR #272, opened 2026-08-03
fix(scope): filter ItemCategory by release window in open_keys lookup

### LINKED ISSUE #271: fix(dep-modules): DORA-1.1.0 tables emit unprefixed open_key duplicates of q-prefixed ones
### Initial Checks

### Summary

Five tables in module `DORA-1.1.0` emit an open_key entry without the `q` prefix in addition to the correct `q`-prefixed one. Pydpm reference only carries the `q`-prefixed variant, so dpmcore is over-declaring.

  | Table    | Extra key in dpmcore | Correct key kept |
  |----------|----------------------|------------------|
  | B_02.02  | LES                  | qLES             |
  | B_03.01  | LEA                  | qLEA             |
  | B_03.03  | LEB                  | qLEB             |
  | B_04.01  | LES                  | qLES             |
  | B_06.01  | LES                  | qLES             |

  The duplicate always mirrors a `q`-prefixed key that is already declared, so it looks like the resolution pass adds an unprefixed alias in addition to the canonical property code.

### Reproducible Example

```python

  engine = create_engine("sqlite:///dpm_4.2.1_20260624.db")
      result = svc.script(module_code="DORA", module_version="1.1.0")
      root = next(iter(result["enriched_ast"].values()))
      print(sorted((root["tables"]["B_02.02"]["open_keys"] or {}).keys()))
  # Actual:   ['CRZ', 'ICT', 'IOB', 'LES', 'TYA', 'qLES', 'qNMP', 'qNMR', 'qNNB']
  # Expected: ['CRZ', 'ICT', 'IOB',        'TYA', 'qLES', 'qNMP', 'qNMR', 'qNNB']
```

### Expected vs Actual

Expected

  Each of the five affected tables emits its open_keys map without the unprefixed alias, matching the pydpm reference (LES/LEA/LEB absent when the q-prefixed version is present).

  Actual

  The five tables carry both the q-prefixed key and its unprefixed alias, inflating the open_keys map. Downstream consumers that look up by exact key may see two entries for what is  conceptually one property.

### dpmcore version

0.1.2rc6

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

victorprietosandamaso-cloud:
Closes #271.

  `get_open_keys_for_tables` filtered `TableVersion` rows by the requested release_id but left `ItemCategory` unfiltered. When a property is renamed across releases (e.g. `LES` valid up to
  release 3, `qLES` valid from release 3 onwards), both `ItemCategory` rows point to the same `ItemID`, so the DISTINCT query returned both codes and every affected open-key ended up
  declared twice in the emitted `open_keys` map.

  On DORA-1.1.0 five tables carried the duplicates: B_02.02, B_03.01, B_03.03, B_04.01 and B_06.01 each emitted both the current q-prefixed code and its historical unprefixed alias. B_02.02
  also carried a stray `TIM` whose ItemCategory window ended before the requested release; the fix cleans that up too.

  Add the same start/end release_id window filter on `ItemCategory` that `TableVersion` already had. Any ItemCategory row outside the requested release window is dropped, so only the
  currently-valid code for each property is emitted. Verified end-to-end against the pydpm reference on DORA-1.1.0 — the five affected tables now match exactly.

  205/205 service unit tests pass; ruff clean.

### DIFF
--- a/src/dpmcore/services/_open_keys.py
+++ b/src/dpmcore/services/_open_keys.py
@@ -65,23 +65,46 @@ def get_open_keys_for_tables(
     )
 
     if release_id is not None:
+        # ``ReleaseID`` values are opaque from DPM 4.2.1 onwards — 4.2.1
+        # is ``1010000003`` while older releases stay in 1..5, and the
+        # transitional ``Playground`` release has an ID larger than
+        # 4.2.1's despite predating it. Release-range comparisons must
+        # therefore go through the date-based sort order in
+        # :mod:`dpmcore.orm.release_sort_order` rather than compare the
+        # numeric IDs directly — a numeric filter happens to give the
+        # right answer for a monotonic ID sequence but silently returns
+        # the wrong window when a release lands out of numeric order.
+        from dpmcore.orm.release_sort_order import (
+            load_release_sort_orders,
+            release_ids_for_sort_order,
+        )
+
+        sort_orders = load_release_sort_orders(session)
+        target_sort = sort_orders.get(release_id)
+        if target_sort is None:
+            raise ValueError(
+                f"release {release_id} has no sort_order — "
+                "no Release row matches that ID."
+            )
+        start_ids = release_ids_for_sort_order(sort_orders, le=target_sort)
+        end_ids = release_ids_for_sort_order(sort_orders, gt=target_sort)
         query = query.filter(
+            TableVersion.start_release_id.in_(start_ids),
             or_(
                 TableVersion.end_release_id.is_(None),
-                TableVersion.end_release_id > release_id,
+                TableVersion.end_release_id.in_(end_ids),
             ),
-            TableVersion.start_release_id <= release_id,
             # ItemCategory has its own release window: a property can
             # be renamed across releases (e.g. ``LES`` up to release 3,
             # ``qLES`` from release 3 onwards) and both rows share the
             # same ``ItemID``. Without this filter both codes end up in
             # the open_keys map for any release, duplicating each
             # property with its historical alias.
+            ItemCategory.start_release_id.in_(start_ids),
             or_(
                 ItemCategory.end_release_id.is_(None),
-                ItemCategory.end_release_id > release_id,
+                ItemCategory.end_release_id.in_(end_ids),
             ),
-            ItemCategory.start_release_id <= release_id,
         )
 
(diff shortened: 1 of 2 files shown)


src/dpmcore/services/_open_keys.py
```diff
@@ -71,6 +71,17 @@ def get_open_keys_for_tables(
                 TableVersion.end_release_id > release_id,
             ),
             TableVersion.start_release_id <= release_id,
+            # ItemCategory has its own release window: a property can
+            # be renamed across releases (e.g. ``LES`` up to release 3,
+            # ``qLES`` from release 3 onwards) and both rows share the
+            # same ``ItemID``. Without this filter both codes end up in
+            # the open_keys map for any release, duplicating each
+            # property with its historical alias.
+            or_(
```
andres-sole, line 104: These  compare ReleaseID values numerically, release-range comparisons cannot rely on numeric ID ordering.

src/dpmcore/services/_open_keys.py
```diff
@@ -71,6 +71,17 @@ def get_open_keys_for_tables(
...
+                ItemCategory.end_release_id.is_(None),
+                ItemCategory.end_release_id > release_id,
+            ),
+            ItemCategory.start_release_id <= release_id,
```
andres-sole: Both new predicates sit inside the if release_id is not None. We need to cover the Release=None case

src/dpmcore/services/_open_keys.py
```diff
@@ -71,6 +71,17 @@ def get_open_keys_for_tables(
...
         )
```
andres-sole, line 117: Add a regression test if possible

andres-sole (changes requested):

src/dpmcore/services/_open_keys.py
```diff
@@ -65,12 +65,46 @@ def get_open_keys_for_tables(
     )
 
     if release_id is not None:
```
andres-sole, line 67: No release means no dedupe, this still returns both LES and qLES for B_02.02. Add an else branch with `filter_active_only(query, ItemCategory.end_release_id)`, same as `get_all_item_signatures`

andres-sole (approved):

PR #276, opened 2026-08-03
fix(scope): exclude tables shared with the home module from dep_modules

### LINKED ISSUE #275: fix(scope): residual over/under-declaration in dep_modules after #253
### Initial Checks

### Summary

#253 narrowed each cross-instance dependency module from the wholesale emission to the referenced subset, but the parity check against the pydpm reference still surfaces small
  over-declaration in 19 modules and small under-declaration in 8. The magnitudes are much smaller than pre-#253 (max +13 tables here vs +130 before), but they are systematic.

  ## Over-declared modules (19)

  Extra tables / URI-level variables / per-table variables that dpmcore emits and the reference does not:

  | Module                       | +tables | +URI-vars | +table-vars |
  |------------------------------|--------:|----------:|------------:|
  | IF_CLASS2-1.3.0              |      13 |        36 |           0 |
  | IF_CLASS2-1.2.0              |      11 |       116 |          12 |
  | IF_CLASS2-1.4.0              |       4 |         0 |           0 |
  | REM_HE_CI-2.2.0              |       2 |       201 |         155 |
  | REM_HE_IF-2.2.0              |       2 |       189 |          56 |
  | COREP_OF-3.1.0               |       2 |        57 |           5 |
  | AE-1.4.0                     |       1 |        80 |          64 |
  | COREP_LR-3.1.0               |       1 |        57 |          16 |
  | AE-1.2.0                     |       1 |        25 |          13 |
  | REM_HR_Institution-2.0.0     |       1 |        10 |           0 |
  | GSII-1.3.0                   |       1 |         2 |           0 |
  | IF_CLASS3-1.4.0              |       1 |         0 |           0 |
  | IF_CLASS3-1.2.0              |       1 |         0 |           0 |
  | FINREP9-3.3.0                |       0 |        82 |          14 |
  | FINREP9DP-1.1.0              |       0 |        79 |          12 |
  | FINREP9-3.1.0                |       0 |        27 |          10 |
  | REM_BM_CI-2.2.0              |       0 |        12 |           6 |
  | REM_BM-2.0.0                 |       0 |        10 |           1 |
  | FP-2.1.0                     |       0 |         2 |           1 |

  ## Under-declared modules (8)
 Variables the reference declares that dpmcore misses. All are small (0 tables missing; only table-level variables):

  | Module              | -URI-vars | -table-vars |
  |---------------------|----------:|------------:|
  | AE-1.2.0            |         0 |          30 |
  | AE-1.4.0            |         0 |          30 |
  | FP-2.1.0            |         0 |           8 |
  | FP-2.2.1            |         0 |           8 |
  | SBP_CR-1.3.0        |         0 |           2 |
  | SBP_CR-1.5.0        |         0 |           0 |
  | COREP_OF-4.1.0      |         1 |           1 |
  | IF_CLASS2-1.4.0     |         0 |           1 |

### Reproducible Example

```python

  engine = create_engine("sqlite:///dpm_4.2.1_20260624.db")
      result = svc.script(module_code="IF_CLASS2", module_version="1.3.0")
      root = next(iter(result["enriched_ast"].values()))
      for uri, entry in root["dependency_modules"].items():
          print(uri, "tables:", len(entry.get("tables", {})),
                "vars:", len(entry.get("variables", {})))
  # Emits 13 tables more than the pydpm reference declares.
```

### Expected vs Actual

For every module, `root["dependency_modules"][<uri>].tables` and `.variables` restrict to exactly the tables and datapoints the cross-instance rules of that module reference. The set
  matches the pydpm reference declaration:

  - IF_CLASS2-1.3.0: dependency module emits the reference's narrow subset, not 13 extra tables.
  - AE-1.4.0: `dependency_modules[<finrep>].variables` matches the reference (no extras and no missing).
  - REM_HE_CI-2.2.0: same set of tables and per-table variables as the reference (currently +2 tables, +201 URI-vars, +155 table-vars extra).
  - Under-declared modules (AE-1.2.0/1.4.0, FP-2.1.0/2.2.1, etc.) include the home-operand variables the reference declares under the shared dependency table.

  Actual:
  `dependency_modules[<uri>]` still carries residual over-declaration in 19 modules and residual under-declaration in 8 modules (see tables in Summary). All counts are one-to-two orders of
  magnitude smaller than the pre-#253 output for the same modules, but every module in the affected list emits a `dependency_modules` block that does not match the pydpm reference exactly.

### dpmcore version

0.1.2rc6

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

victorprietosandamaso-cloud:
Closes #275 partially — fixes the shared-table subset; residual over-declaration on IF_CLASS2 traced to a separate over-inclusion of operations, filed undera follow-up.
 
 `_build_dependency_entry` iterated `ModuleVersionComposition` to enumerate the tables of each dependency module, keeping every one that carries variables.
  When a table is declared by both the primary (home) module and a dependency — a shared table like `I_05.00`, present in both IF_CLASS2 and IF_CLASS3 — the sharing belongs to the home declaration, not to the dependency, so the `dep_modules` block ended up naming the same table on both sides.

  `ASTGeneratorService._build_dependency_info` now computes the primary module's table set once per script (the caller loops per operation with a fixed  `primary_module_vid`) and threads it into `ScopeCalculatorService.detect_cross_module_dependencies` as `home_module_tables`. The callee falls back to
  `_get_module_tables(primary_module_vid, release_id=release_id).keys()` when the parameter is not supplied, so direct callers keep working unchanged.

  Inside `_build_dependency_entry` the exclusion runs after the `ref_period` loop and after `_narrow_dependency_tables`:
 
  - `ref_period` is computed on the full `tables_dict` so a time-shifted table isn't dropped before its `T-1Q` (or equivalent) is captured.
  - Narrowing (#250) still sees every operand-referenced table, so a validation whose operand set is entirely inside a shared table narrows correctly instead
  of triggering the narrowing empty-fallback and reintroducing the shared table via the module-wide unnarrowed set.
  - The exclusion is unconditional — shared tables are stripped even when the exclusion would empty `tables_dict`. A full-overlap dep still surfaces in `cross_instance_dependencies` via its `URI`; #251 resolves cross-instance operands off `referenced_variables` when the caller supplies them.

Empirically verified against the pydpm reference on the fixture DB:
 
  - IF_CLASS3-1.2.0: extra tables 1 → 0 (fully clean).
  - IF_CLASS3-1.4.0: 1 → 0.
  - IF_CLASS2-1.4.0: 4 → 0.
  - IF_CLASS2-1.2.0: 11 → 4 (partial).
  - IF_CLASS2-1.3.0: 13 → 5 (partial).

  The residual +4 / +5 on IF_CLASS2 sit on the `corep_of` dependency URI, not on a shared home table. Investigation shows the extras are dragged in by operations (`v09813_m`, `v09812_m`, `v7380_m`, `v7381_m`, `v09814_m`) that dpmcore includes in IF_CLASS2-1.3.0 but the pydpm reference does not — these ops reference only COREP_OF tables and pydpm does not scope them to IF_CLASS2 at all. That is the shape of the `finding1_dpmcore_extra_ops` divergence (41modules across the parity run) and lives in `filter_valid_dependency_modules` / `is_cross_module`, not in `_narrow_dependency_tables`. Filed as a separate follow-up.
  
 Regression tests added under a new `tests/unit/services/test_scope_calculator_shared_home.py`: shared-table exclusion, the narrow-then-exclude order, full-overlap with and without `referenced_variables`, and the `home_module_tables` precomputed-parameter contract. Ruff + mypy clean

### DIFF
--- a/src/dpmcore/services/ast_generator.py
+++ b/src/dpmcore/services/ast_generator.py
@@ -14,6 +14,7 @@ from typing import (
     Iterable,
     List,
     Optional,
+    Set,
     Tuple,
     Union,
 )
@@ -1271,6 +1272,17 @@ class ASTGeneratorService:
         all_dep_modules: Dict[str, Any] = {}
         all_scope_results: List["ScopeResult"] = []
 
+        # The home-module table set is a per-script constant (the
+        # primary module never changes across this loop), and computing
+        # it inside ``detect_cross_module_dependencies`` would repeat a
+        # per-table variable/open-key fetch on every iteration. Compute
+        # it once here and thread it through.
+        home_module_tables: Set[str] = set(
+            self._scope_calc._get_module_tables(
+                primary_module_vid, release_id=release_id
+            ).keys()
+        )
+
         for item, sr, ts, refs in scope_pairs:
             all_scope_results.append(sr)
             op_code = item[1]
@@ -1283,6 +1295,7 @@ class ASTGeneratorService:
                 compute_alternative_deps=False,
                 referenced_variables=refs.variables,
                 referenced_tables=refs.tables,
+                home_module_tables=home_module_tables,
             )
             all_intra.extend(current.get("intra_instance_validations", []))
             self._merge_cross_deps(
--- a/src/dpmcore/services/scope_calculator.py
+++ b/src/dpmcore/services/scope_calculator.py
@@ -239,6 +239,7 @@ class ScopeCalculatorService:
         release_code: Optional[str] = None,
         referenced_variables: Optional[Dict[str, str]] = None,
         referenced_tables: Optional[Set[str]] = None,
+        home_module_tables: Optional[Set[str]] = None,
     ) -> Dict[str, Any]:
         """Build dependency information for a scope result.
 
@@ -267,6 +268,14 @@ class ScopeCalculatorService:
                 narrows each dependency module's declaration to the
                 subset the operation uses (#250); omit both to declare
                 the dependency modules whole.
+            home_module_tables: Optional pre-computed set of table codes
+                owned by the primary (home) module — the same set this
+                method would derive from :meth:`_get_module_tables`.
+                Callers that iterate this method with a fixed
+                ``primary_module_vid`` (per-op dependency detection
+                over a script's operations) should pass a single
+                pre-computed set to avoid re-running the per-table
+                variable/open-key fetch on every iteration.
 
         Returns a dict with:
         - ``intra_instance_validations``
@@ -350,11 +359,18 @@ class ScopeCalculatorService:
         # exclude them from dependency_modules[<dep>].tables. Without
         # this, cross-module ops that touch a shared table declare it
         # twice (once in home, once in every dep that also owns it).
-        primary_tables = set(
-            self._get_module_tables(
-                primary_module_vid, release_id=release_id
-            ).keys()
-        )
+        # Caller-supplied ``home_module_tables`` avoids the per-op
+        # recompute when this method runs inside a loop with a fixed
+        # ``primary_module_vid`` (the query is a per-table variable/
+        # open-key fetch, not a code-only lookup).
+        if home_module_tables is None:
+            primary_tables = set(
+                self._get_module_tables(
+                    primary_module_vid, release_id=release_id
+                ).keys()
+            )
+        else:
+            primary_tables = home_module_tables
 
         for vid in sorted_vids:
             mv = mv_by_vid.get(vid)
@@ -432,21 +448,6 @@ class ScopeCalculatorService:
         }
         if not tables_dict:
             return None
-        # Drop tables the primary (home) module also declares — the shared
-        # ones belong to the home declaration. Skip if the exclusion would
-        # leave the dep empty (a dep whose entire table set is contained in
-        # the home is either a genuine full overlap we still want to declare
-        # or a test fixture where the mock returns the same tables for
-        # every vid).
-        home = home_module_tables or set()
-        if home:
-            filtered = {
-                tcode: tdata
-                for tcode, tdata in tables_dict.items()
-                if tcode not in home
-            }
-            if filtered:
-                tables_dict = filtered
 
         # The timeshift is a module-level property carried by the dependency
         # module's tables. Compute it BEFORE narrowing: narrowing drops any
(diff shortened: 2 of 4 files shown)


src/dpmcore/services/scope_calculator.py
```diff
@@ -344,6 +344,18 @@ def detect_cross_module_dependencies(
...
+        # also lists any of these is sharing a table with the home; the
+        # sharing belongs to the home declaration, not to the dep, so
+        # exclude them from dependency_modules[<dep>].tables. Without
+        # this, cross-module ops that touch a shared table declare it
+        # twice (once in home, once in every dep that also owns it).
+        primary_tables = set(
+            self._get_module_tables(
+                primary_module_vid, release_id=release_id
+            ).keys()
+        )
```
ruizmaa: `_get_module_tables(primary_module_vid, ...)` is a heavy query which runs once per operation, not once per script: `detect_cross_module_dependencies` is called inside a loop in `ast_generator.py`'s `_build_dependency_info`, with the same `primary_module_vid` every time:

    // ast_generator.py:1274-1281
    for item, sr, ts, refs in scope_pairs:
        current = self._scope_calc.detect_cross_module_dependencies(
            scope_result=sr,
            primary_module_vid=primary_module_vid,   // same value every iteration
            operation_code=op_code,
            release_id=release_id,
            ...
        )

`scope_pairs` has one entry per cross-module operation, so this repeats the same query N times for an identical result. Also worth noting, `ast_generator.py` already computes this a few lines earlier for a different purpose:

    // ast_generator.py:266-268
    primary_tables_full = self._scope_calc._get_module_tables(
        primary_module_vid, release_id=release_id
    )

Could `home_module_tables` be threaded through from there instead of recomputed here?

ruizmaa (commented):
New tests should be added to test this new behaviour

src/dpmcore/services/scope_calculator.py
```diff
@@ -410,6 +432,21 @@ def _build_dependency_entry(
...
+        # every vid).
+        home = home_module_tables or set()
+        if home:
+            filtered = {
+                tcode: tdata
+                for tcode, tdata in tables_dict.items()
+                if tcode not in home
+            }
+            if filtered:
+                tables_dict = filtered
```
ruizmaa: This guard looks wrong to me. If excluding shared tables would leave the dependency empty, it skips the exclusion entirely, reintroducing the exact duplicate-table bug this PR fixes.

Also I don't follow the "or a test fixture where the mock returns the same tables for every vid" part of the comment. If that's the real reason, I think we should think another way to build those fixtures

andres-sole (commented):
The exclusion is in the wrong place. In `_build_dependency_entry` it runs at line 465, before the `ref_period` loop and before `_narrow_dependency_tables`, so both of those end up working on a table set that's missing the tables the operation actually references.

I tried it on the fixture DB with IF_CLASS2-1.3.0 - COREP_OF-4.0.0, for an op referencing only the shared `C_18.00`. The exclusion drops `C_18.00`, narrowing then cant find it and returns empty, and the fallback puts the whole module back. Moving the exclusion after the ref_period loop and after narrowin fixes all that

src/dpmcore/services/ast_generator.py
```diff
@@ -1281,6 +1272,25 @@ def _build_dependency_info(
...
+        # it once here and thread it through. Fall back to a lazy
+        # in-callee compute when ``_scope_calc`` is a mock that can't
+        # answer this eagerly (test stubs).
+        try:
+            home_module_tables: Optional[Set[str]] = set(
+                self._scope_calc._get_module_tables(
+                    primary_module_vid, release_id=release_id
+                ).keys()
+            )
+        except TypeError:
```
andres-sole: Unreachabe code

src/dpmcore/services/scope_calculator.py
```diff
@@ -669,6 +469,35 @@ def _build_dependency_entry(
...
+        # set: over-declaring the shared tables is wrong, but silently
+        # dropping a genuine cross-instance dependency is worse (mirrors
+        # the ``_narrow_dependency_tables`` fallback contract).
+        if home_module_tables:
+            filtered = {
+                tcode: tdata
+                for tcode, tdata in tables_dict.items()
+                if tcode not in home_module_tables
+            }
+            if filtered or referenced_variables:
```
andres-sole: the guard is relaxed, not resolved. @ruizmaa was talking of full-overlap with no `referenced_variables`. the shared tables stay

andres-sole (approved):

Issue #121, opened 2026-06-08
Should yield semantic error when results are not expected


antonio-olleros:
### Initial Checks

- [x] I have searched [existing issues](https://github.com/Meaningful-Data/dpmcore/issues?q=is%3Aissue) for duplicates

### Problem Description

Please check Validation v6200_m (last version). There are wheres with different values, and therefore the innerjoin will never yield results.

### Proposed Solution

```python
Yield semantic error
```

PR #280, opened 2026-08-03
feat(dpm-xl): accept the Fact Component in a where clause

### LINKED ISSUE #266: feat(dpm-xl): extend clause operators (where/get/rename) to fact and attribute components
### Initial Checks

### Problem Description

The clause operators `[where …]`, `[get …]` and `[rename …]` currently accept only DPM key components (open-keys) as targets. Any expression that filters, projects or renames on the Fact Component (`f`) or on a declared Attribute Component of the recordset is rejected by the semantic analyzer with `SemanticError 1-5 "open keys not found: [...]"`.

  This blocks a family of checks that filter a recordset by its fact value or by an attribute. The most immediate examples are EGDQ_0070, EGDQ_0071a, EGDQ_0071b and EGDQ_0133, all of which need `[where f = <item>]` on the operand recordset.

  The DPM-XL specification has been extended to allow these usages. Grammar changes are not needed on the dpmcore side — the existing lexer already accepts the identifiers; the work is in the semantic analyzer and in the JSON payload emitted by `ASTGeneratorService.script`.

Reproducible example (blocked today):

  ```python
  from dpmcore.services.semantic import SemanticService
  # expr = (
  #   "{tC_06.02, c0030}[where f and LGS != entityID][get LGS] in "
  #   "set_of({tC_40.01, c0095}[where f in {[eba_CT:x12], [eba_CT:x18]}][get LIN])"
  # )
  # Actual:   SemanticError 1-5 "open keys not found: ['f']".
  # Expected: passes semantic once the extension ships.


### Proposed Solution

```python
Extend the semantic analyzer so that clause operators resolve identifiers against Fact and Attribute Components in addition to DPM key components.

  Scope, in priority order:

  1. [where <condition>] — accept conditions referencing any Component of the operand (DPM Key Components already supported, Fact Component new, Attribute Components new). Standard Key
  Components (r, c, s) remain disallowed inside where. Highest-priority sub-item — unblocks EGDQ_0070, EGDQ_0071a, EGDQ_0071b, EGDQ_0133.
  2. [get <component>] — extend to project on Attribute Components. Projection on the Fact Component remains disallowed.
  3. [rename <component>] — extend to rename Attribute Components. Renaming the Fact Component remains disallowed.
  4. sub clause — unchanged; the existing restriction to Key Components should be stated explicitly.
  5. The result Fact Component of the operation takes the Data Type of the selected Component.
  6. where_expression inside a with clause referencing only the Fact Component is applicable to selections on tables.

  The AST node representation and the shape of the JSON payload emitted for the new constructs need to be aligned with Adam Engine and CVE, the two downstream consumers of
  ASTGeneratorService.script. Options to weigh before implementation:

  - Dedicated AST node classes (FactComponentRef, AttributeComponentRef) vs. a flag on the existing Dimension node.
  - Whether Attribute names resolve against a reserved list or against any attribute declared in the recordset structure.
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Closes #266 

The clause operators accepted only DPM key components as targets, so any `where` condition filtering a recordset by its fact value was rejected by the semantic analyzer with `1-5 "open keys not found: ['f']"`. The dictionary pre-check resolves every clause identifier as an open key, and the Fact Component has no dictionary row. That blocked EGDQ_0070, EGDQ_0071a, EGDQ_0071b and EGDQ_0133.

Following the spec extension, `[where …]` now accepts the Fact Component. Concretely:

- **`where` accepts `f`.** The Fact is excluded from the open-key lookups and resolved structurally instead, against the Fact Component of the enclosing clause operand — so it is typed per operand. `[where f]` is a valid filter on a Boolean-typed selection and a type error (`3-3`) on a numeric one, and `[where f = …]` type-checks against the operand's data type.
- **`get`, `rename` and `sub` keep rejecting `f`,** but now with the structural error `4-5-0-1` ("clause operators can't be used with standard key components or fact component") rather than the misleading `1-5`. Standard key components (`r`, `c`, `s`) remain disallowed for every clause operator, and the `sub` clause gained the guard it was missing, so a backticked ``[sub `r` = …]`` reports `4-5-0-1` instead of falling through to `2-8`.
- **The Fact never contributes a where-equality pin.** Pins exist to detect an inner join that can never match (#121) and that join is over key components; two operands filtering on different fact values are perfectly joinable, so pinning the Fact would have produced a false `2-2`.
- **The result Fact Component of a `get` already took the data type of the selected component** — that part of the extension needed no change, only a test.

No grammar change was needed: `f` already lexes as an ordinary property code inside a clause, so the rejection was always semantic.

## Wire format

Unchanged, and no engine schema change is required. The Fact travels as `{"class_name": "Dimension", "dimension_code": "f"}`, exactly like a key dimension. The engine's `Dimension` definition accepts precisely `class_name` + `dimension_code` with `additionalProperties: false` and a closed `class_name` enum, so a dedicated node class or an extra role field would both have been rejected; component names are unique within a recordset and the Fact is named `f` by definition, so the code resolves unambiguously by name.

## Scope

This delivers the `where`-on-Fact part of the extension. Two pieces are deliberately left out, each with a follow-up issue drafted:

- `get` / `rename` over Attribute Components. `AttributeComponent` and the whole `propagate_attributes` machinery exist in the symbol layer, but nothing ever constructs an attribute component, and the current dictionary ships no attribute variables — so there is nothing to resolve against. Once a producer exists, the shared clause validation already resolves names against attributes, so most of that behaviour follows for free.
- A Fact-only `where_expression` in a `with` clause. The grammar accepts the `with … [where …]` block but the AST drops it, so the clause applies to nothing today and no error is raised. That is a separate defect, and the "must have at least one key component" constraint belongs with it.

## Checklist

- [x] Tests pass (`pytest`)
- [x] Documentation updated (clause-operator target matrix in `specification/03-services.md`, which was also missing `GET` from the operator list)

## Notes

### DIFF
--- a/tests/integration/validation/test_semantic_where_fact.py
+++ b/tests/integration/validation/test_semantic_where_fact.py
@@ -1,0 +1,95 @@
+"""Issue #266 — ``[where f …]`` against the real dictionary.
+
+Before the extension, a ``where`` condition naming the Fact Component died in
+the operands pre-check with ``1-5 "open keys not found: ['f']"``, because
+``f`` has no dictionary row and every condition identifier was resolved as an
+open key. It must now resolve structurally, against the Fact Component of the
+clause operand, while ``get``/``rename``/``sub`` keep rejecting it — with
+``4-5-0-1`` rather than the misleading ``1-5``.
+"""
+
+from dpmcore.services.semantic import SemanticService
+
+RELEASE = "4.2.1"
+
+# C_08.01.a r0020 c0260 is the operand of published check v6200_m: a
+# monetary cell whose selection carries an open key (``qEEA``).
+OPERAND = "{tC_08.01.a, r0020, c0260}"
+
+
+def test_where_on_fact_is_valid(fixture_session):
+    """A Fact-only where condition passes semantic validation."""
+    result = SemanticService(fixture_session).validate(
+        f"{OPERAND}[where f > 0] >= 0", release_code=RELEASE
+    )
+    assert result.is_valid, result.error_message
+
+
+def test_where_mixing_fact_and_open_key_is_valid(fixture_session):
+    """The Fact composes with the DPM key components it used to exclude."""
+    expression = (
+        f"sum({OPERAND}[where f > 0 and qEEA in {{[eba_qAE:qx2018]}}]) >= 0"
+    )
+    result = SemanticService(fixture_session).validate(
+        expression, release_code=RELEASE
+    )
+    assert result.is_valid, result.error_message
+
+
+def test_where_on_fact_then_get_on_a_key_is_valid(fixture_session):
+    """Chained clauses: the outer ``get`` still sees the filtered operand."""
+    result = SemanticService(fixture_session).validate(
+        f"{OPERAND}[where f > 0][get qEEA] >= 0", release_code=RELEASE
+    )
+    assert result.is_valid, result.error_message
+
+
+def test_disjoint_fact_filters_do_not_yield_2_2(fixture_session):
+    """Unlike a key, the Fact cannot make an inner join empty.
+
+    Guards against a false ``2-2`` from the contradictory-where check of
+    issue #121, which is keyed on *Key Components* only.
+    """
+    result = SemanticService(fixture_session).validate(
+        f"{OPERAND}[where f = 1] = {OPERAND}[where f = 2]",
+        release_code=RELEASE,
+    )
+    assert result.is_valid, result.error_message
+
+
+def test_non_boolean_fact_condition_yields_3_3(fixture_session):
+    """``f`` is typed from the operand, so a bare numeric fact is not a filter.
+
+    This is what makes ``[where f and …]`` meaningful only on Boolean-typed
+    selections: the condition must promote to Boolean.
+    """
+    result = SemanticService(fixture_session).validate(
+        f"{OPERAND}[where f] >= 0", release_code=RELEASE
+    )
+    assert not result.is_valid
+    assert result.error_code == "3-3"
+
+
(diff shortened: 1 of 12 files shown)


victorprietosandamaso-cloud (commented):
The spec states that a where referencing only the Fact Component must be valid on any selection, including tables with no open keys. The code covers this implicitly by excluding f from the
  open-key lookup, but no test asserts it. Worth adding a regression test on a selection over a table without open keys to guard against future changes to that lookup silently breaking it.

tests/integration/validation/test_semantic_where_fact.py
```diff
@@ -0,0 +1,124 @@
...
+    result = SemanticService(fixture_session).validate(
+        expression, release_code=RELEASE
+    )
+    assert result.is_valid, result.error_message
+
+
+def test_where_on_fact_then_get_on_a_key_is_valid(fixture_session):
+    """Chained clauses: the outer ``get`` still sees the filtered operand."""
+    result = SemanticService(fixture_session).validate(
+        f"{OPERAND}[where f > 0][get qEEA] >= 0", release_code=RELEASE
```
ruizmaa, line 71: Could we also add the reverse order, `[get qEEA][where f > 0]`? Get re-types the result Fact Component to the selected component's type (`ClauseOperator.generate_result_structure`), so it'd be good to have an explicit test that a subsequent `where f` resolves against that re-typed Fact correctly.

andres-sole: Done in latest changes

ruizmaa (approved):

PR #290, opened 2026-08-05
Reject duplicate persistent-assignment targets in a script

### LINKED ISSUE #289: Semantic validation doesn't flag two independent formulas assigning the same variable for the same reference period
### Initial Checks

### Summary

A DPM-XL variable can only have one value per reference period. Two cells that resolve to the same variable should only be allowed to coexist if one is a plain equality alias of the other (`{B} <- {A}`). Two independent formulas targeting the same variable should be rejected, since there is no way to reconcile two different results for one value.

### Reproducible Example

```python
from dpmcore.services.semantic import SemanticService

engine = create_engine("sqlite:///tests/fixtures/test_data.db")

svc = SemanticService(session)

# Two independent, conflicting formulas assigning the same cell/variable
script = "{tF_01.01, r0010, c0010} <- 1; {tF_01.01, r0010, c0010} <- 2;"
result = svc.validate(script, release_code="4.2.1")

print(result.is_valid)      # True, should be False
print(result.error_code)    # None, should identify the conflicting assignment

# is_valid: True
# error_code: None
# error_message: None
# warning: None

# An equality-alias pair (`{tF_01.01, r0010, c0010} <- 1; {tF_01.01, r0020, c0010} <- {tF_01.01, r0010, c0010};`) should remain valid, only the independent/conflicting case is the problem
```

### Expected vs Actual

Expected: `result.is_valid` is `False`, with a semantic error identifying the variable that more than one independent formula assigns. A plain equality alias of an existing assignment (`{B} <- {A}`) should still validate as `True`.

Actual: `result.is_valid` is `True`, dpmcore silently accepts a script with two unrelated, conflicting formulas for the same variable, with no error and no warning.

### dpmcore version

0.1.2rc6

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #289 

## Summary

Semantic validation accepted scripts with two independent, conflicting formulas assigning the same `{cellRef}`/`{varRef}` (e.g. `{tF_01.01, r0010, c0010} <- 1; {tF_01.01, r0010, c0010} <- 2;`), even though a DPM-XL variable can only hold one value per reference period.

Added a check in `InputAnalyzer.visit_Start` that raises error `6-1` when two top-level statements target the same output. A dependency between two different cells (`{B} <- {A}`) is unaffected.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/semantic_analyzer.py
+++ b/src/dpmcore/dpm_xl/semantic_analyzer.py
@@ -118,7 +118,7 @@ _PARAMETER_SCALAR_TYPES: dict[str, str] = {
 
 def _check_duplicate_persistent_assignments(children: list[AST]) -> None:
     """Raise ``6-1`` when two statements assign the same ``{cellRef}``/``{varRef}``."""
-    seen: dict[tuple[Any, ...], str] = {}
+    seen: set[tuple[Any, ...]] = set()
     for child in children:
         target = child
         if isinstance(target, TemporaryAssignment):
@@ -153,7 +153,7 @@ def _check_duplicate_persistent_assignments(children: list[AST]) -> None:
             continue
         if key in seen:
             raise errors.SemanticError("6-1", variable=variable)
-        seen[key] = variable
+        seen.add(key)
 
 
 class InputAnalyzer(ASTTemplate, ABC):
--- a/src/dpmcore/dpm_xl/semantic_analyzer.py
+++ b/src/dpmcore/dpm_xl/semantic_analyzer.py
@@ -2,6 +2,7 @@ from abc import ABC
 from typing import Any, cast
 
 import pandas as pd
+from sqlalchemy.orm import Session
 
 from dpmcore import errors
 from dpmcore.dpm_xl.ast.nodes import (
@@ -50,6 +51,7 @@ from dpmcore.dpm_xl.ast.where_clause import (
     collect_where_equality_pins,
     merge_where_constraints,
 )
+from dpmcore.dpm_xl.model_queries import ViewDatapointsQuery
 from dpmcore.dpm_xl.operators.clause import Sub as SubOperator
 from dpmcore.dpm_xl.symbols import (
     Component,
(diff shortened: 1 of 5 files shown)


src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -115,6 +116,46 @@
...
+    seen: dict[tuple[Any, ...], str] = {}
+    for child in children:
+        target = child
+        if isinstance(target, TemporaryAssignment):
+            target = target.right
+        if not isinstance(target, PersistentAssignment):
+            continue
+        left = target.left
+        if isinstance(left, VarID):
+            key: tuple[Any, ...] = (
```
andres-sole: When the target has no table of its own, the key is just rows + cols, so these two collide:

```
{r0010, c0010} <- with {tF_01.01}: {r0020, c0010};
{r0010, c0010} <- with {tF_01.02}: {r0020, c0010};
```

Each statement validates on its own, but together they get 6-1. They write to different tables, so this shouldn't be a duplicate.

src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -115,6 +116,46 @@
...
+                "VarID",
+                left.table,
+                left.operation,
+                left.is_table_group,
+                tuple(sorted(left.rows)) if left.rows else None,
```
andres-sole: Range or a wildcard never matches the codes it covers. These pass today:

```
{tF_01.01, r*, c0010} <- 1; {tF_01.01, r0010, c0010} <- 2;
{tF_01.01, r0010-r0030, c0010} <- 1; {tF_01.01, (r0010,r0020,r0030), c0010} <- 2;
```

andres-sole: `tuple(sorted(...))` keeps duplicates, so `(r0010, r0010)` and `r0010` give different keys for the same cell. `frozenset(left.rows)` covers the ordering and the repetition at once.

src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -115,6 +116,46 @@
...
+            target = target.right
+        if not isinstance(target, PersistentAssignment):
+            continue
+        left = target.left
+        if isinstance(left, VarID):
+            key: tuple[Any, ...] = (
+                "VarID",
+                left.table,
+                left.operation,
+                left.is_table_group,
```
andres-sole: With `is_table_group` in the key, a group target and one of its member tables are never compared

src/dpmcore/dpm_xl/semantic_analyzer.py
```diff
@@ -115,6 +116,46 @@
 }
 
 
+def _check_duplicate_persistent_assignments(children: list[AST]) -> None:
+    """Raise ``6-1`` when two statements assign the same ``{cellRef}``/``{varRef}``."""
+    seen: dict[tuple[Any, ...], str] = {}
```
andres-sole: `seen` is never read

andres-sole (changes requested):

andres-sole (approved):

PR #296, opened 2026-08-10
Fix releases: treat "playground" NOT NULL sentinel and self-reference as latest

### LINKED ISSUE #295: Release-range filtering treats a dated "always open" sentinel release as the oldest release, not the latest
### Initial Checks

### Summary

`release_sort_order.py` treats a release with `date IS NULL` as an "always open" release. DPM Studio schemas declare `Release.Date` `NOT NULL`, so they can't leave it empty: their "always open" release (`Release.Type = "playground"`) is stored with a placeholder date instead (e.g. `1970-01-01`). dpmcore has no way to know that date is a placeholder, so it compares it like any other real date. And a decades-old placeholder date compares as an old, long-past release, not an open one. So any row pointing its `EndReleaseID` at it looks like it closed back in 1970, and disappears from queries at the real current release.

### Reproducible Example

```python
from datetime import date

from dpmcore.dpm_xl.model_queries import ModuleVersionQuery
from dpmcore.dpm_xl.utils.filters import filter_by_release
from dpmcore.orm import Base
from dpmcore.orm.rendering import Table, TableVersion

    session.add_all(
        [
            Release(release_id=1, code="4.2", date=date(2025, 10, 31)),
            Release(release_id=2, code="4.2.1", date=date(2026, 2, 15)),
            # "Always open" sentinel, with a placeholder date instead of NULL.
            Release(
                release_id=9999,
                code="Playground",
                date=date(1970, 1, 1),
                type="playground",
            ),
            Table(table_id=1),
            # Still open today, via the sentinel (end_release_id=9999).
            TableVersion(
                table_vid=1,
                table_id=1,
                code="T1",
                start_release_id=1,
                end_release_id=9999,
            ),
        ]
    )
    session.commit()

    print("last release:", ModuleVersionQuery.get_last_release(session))

    q = session.query(TableVersion)
    filtered = filter_by_release(
        q,
        start_col=TableVersion.start_release_id,
        end_col=TableVersion.end_release_id,
        release_id=2,
    )
    print("T1 visible at release 4.2.1:", [tv.table_vid for tv in filtered.all()])
```

### Expected vs Actual

Expected: `T1 visible at release 4.2.1: [1]` — `T1` was never closed, so it should be visible at any real release.

Actual: `T1 visible at release 4.2.1: []` — `T1` disappears, because `filter_by_release` compares the Playground release's literal `1970-01-01` date and finds nothing sorts after it.

### dpmcore version

0.1.2rc7

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Fixes #295 

## Summary

Schemas with `Release.Date` NOT NULL mark their perpetual "always open" release via `Release.Type = "playground"` instead, with a meaningless placeholder date (e.g. `1970-01-01`). dpmcore compared that date literally, so it ranked as the oldest release instead of the latest. Any row "still open" via that sentinel looked closed, and disappeared from queries at the real current release.

- `compute_sort_order` (and every caller that loads `Release.date`) now also takes `Release.type`; a type in `_NON_CHRONOLOGICAL_TYPES` (currently just `"playground"`) sorts as latest regardless of its literal date.
- `filter_by_release` now also treats a row ending at any "always latest" release (undated or non-chronological) as open when queried at that same release.
- `filter_item_version` had the identical self-reference gap (same pattern, used by `HierarchyService` for versioned `ItemCategory`/`Item`). Not covered by the original report, but reachable the same way. Fixed identically.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/utils/filters.py
+++ b/src/dpmcore/dpm_xl/utils/filters.py
@@ -208,17 +208,26 @@ def filter_active_only(query: Any, end_col: Any) -> Any:
     """Filter for currently active records.
 
     Args:
-        query: SQLAlchemy Query or Select object.
+        query: SQLAlchemy ``Query`` — must be a session-bound ``Query``
+            (i.e. produced by ``Session.query(...)``); see
+            :func:`filter_by_release`.
         end_col: Column for end release.
 
     Returns:
         Filtered query: ``end_col IS NULL``, or ``end_col`` is itself an
         "always latest" release (undated or non-chronological), which
         never actually closes.
+
+    Raises:
+        TypeError: If ``query`` is not a session-bound SQLAlchemy
+            ``Query``.
     """
     session = getattr(query, "session", None)
     if session is None:
-        return query.filter(end_col.is_(None))
+        raise TypeError(
+            "filter_active_only(query=...) expects a session-bound "
+            "SQLAlchemy Query (Session.query(...))."
+        )
     sort_orders = load_release_sort_orders(session)
     perpetual_ids = release_ids_for_sort_order(
         sort_orders, ge=compute_sort_order(None)
--- a/src/dpmcore/dpm_xl/utils/filters.py
+++ b/src/dpmcore/dpm_xl/utils/filters.py
@@ -140,16 +140,18 @@ def filter_by_release(
             or undated release, which always ranks as the latest
             regardless of its literal date.
         If release_id is None:
-            * ``active_only_fallback=True`` → ``end_col IS NULL`` only
-              (currently-active rows). Useful for callers that want a
-              deterministic default when no release is supplied.
-            * Otherwise → return query unmodified.
+            * ``active_only_fallback=True`` → apply
+              :func:`filter_active_only` (see its own docstring for the
+              perpetual-end exception).
+            * Otherwise → return query unmodified, no session required.
 
     Args:
         query: SQLAlchemy ``Query`` — must be a session-bound ``Query``
-            (i.e. produced by ``Session.query(...)``). Core ``Select``
-            statements are not supported because the helper needs the
-            session to load the release mapping.
+            (i.e. produced by ``Session.query(...)``) whenever a session
+            is actually needed (``release_id`` given, or
+            ``active_only_fallback=True``). Core ``Select`` statements
+            are not supported because the helper needs the session to
+            load the release mapping.
         start_col: Column for start release ID (FK to ``Release``).
         end_col: Column for end release ID (FK to ``Release``).
         release_id: Release ID to filter for.
@@ -161,14 +163,13 @@ def filter_by_release(
         Filtered query.
 
     Raises:
-        TypeError: If ``query`` is not a session-bound SQLAlchemy
-            ``Query`` (e.g. a Core ``Select`` was passed).
+        TypeError: If a session is needed (see ``query`` above) and
+            ``query`` is not a session-bound SQLAlchemy ``Query`` (e.g. a
+            Core ``Select`` was passed).
         ValueError: If ``release_id`` does not correspond to a known
             ``Release`` row.
     """
-    if release_id is None:
-        if active_only_fallback:
-            return filter_active_only(query, end_col)
+    if release_id is None and not active_only_fallback:
         return query
 
     session = getattr(query, "session", None)
@@ -179,6 +180,9 @@ def filter_by_release(
             f"{type(query).__name__!r} which has no .session — Core "
             "Select statements are not supported.",
         )
+
+    if release_id is None:
+        return filter_active_only(query, end_col)
 
     # Order releases by date. resolve_sort_order already handles
     # non-chronological types like "Playground". This block only compares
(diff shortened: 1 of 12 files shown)


src/dpmcore/services/structure.py
```diff
@@ -288,21 +289,18 @@ def query_releases(
...
-        # NULL-first CASE keeps this uniform across backends (SQLite and
-        # SQL Server otherwise sort NULLs last in DESC).
-        q = q.order_by(
-            case((Release.date.is_(None), 1), else_=0).desc(),
-            Release.date.desc(),
+        # Latest first via compute_sort_order
+        rows = sorted(
+            q.all(),
+            key=lambda r: (compute_sort_order(r.date, r.type), r.release_id),
+            reverse=True,
```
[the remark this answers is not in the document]
ruizmaa: This table is tiny in practice, so DB-side pagination wouldn't help.

The previous SQL CASE duplicated this exact ordering rule and drifted out of sync with compute_sort_order. In my opinion, keeping one implementation is worth more than optimizing for a scale this table won't reach.

src/dpmcore/orm/release_sort_order.py
```diff
@@ -33,24 +40,35 @@
 # unpublished working release ranks as the latest.
 _UNDATED_SORT_ORDER = date.max.toordinal() + 1
 
+# ``Release.type`` values marking a perpetual release (sorts as latest,
+# like an undated one) on schemas where ``Release.Date`` is ``NOT NULL``.
+_NON_CHRONOLOGICAL_TYPES: FrozenSet[str] = frozenset({"playground"})
```
andres-sole, line 45: Normalise with `str(release_type).strip().lower()`

src/dpmcore/dpm_xl/utils/filters.py
```diff
@@ -200,13 +208,31 @@ def filter_active_only(query: Any, end_col: Any) -> Any:
...
+        never actually closes.
+
+    Raises:
+        TypeError: If ``query`` is not a session-bound SQLAlchemy
+            ``Query``.
     """
-    return query.filter(end_col.is_(None))
+    session = getattr(query, "session", None)
+    if session is None:
+        raise TypeError(
```
andres-sole, line 231: `filter_active_only` now requires a session. `filter_by_release(..., active_only_fallback=True)` inherits the break

src/dpmcore/services/structure.py
```diff
@@ -288,21 +290,18 @@ def query_releases(
...
         )
 
         if latest or latest_stable:
-            q = q.limit(1)
+            rows = rows[:1]
         else:
-            q = q.offset(offset).limit(limit)
```
andres-sole, line 303: `rows[offset : offset + limit]` replaced SQL OFFSET/LIMIT: `limit=None` now raises `TypeError` (was "no limit"), `limit=-1` silently returns wrong rows

src/dpmcore/orm/release_sort_order.py
```diff
@@ -33,24 +40,35 @@
 # unpublished working release ranks as the latest.
 _UNDATED_SORT_ORDER = date.max.toordinal() + 1
```
andres-sole, line 41: every undated *and* every `"playground"`-typed release collapses onto one sentinel. Breaks the "dates are unique, no tiebreak needed"

src/dpmcore/orm/release_sort_order.py
```diff
@@ -33,24 +40,35 @@
...
 
+# ``Release.type`` values marking a perpetual release (sorts as latest,
+# like an undated one) on schemas where ``Release.Date`` is ``NOT NULL``.
+_NON_CHRONOLOGICAL_TYPES: FrozenSet[str] = frozenset({"playground"})
+
 
-def compute_sort_order(release_date: Optional[date]) -> int:
+def compute_sort_order(
+    release_date: Optional[date], release_type: Optional[str] = None
```
andres-sole: `release_type` is optional with default `None`, so `compute_sort_order(r.date)` still type-checks and silently reverts to pre-fix ordering.

andres-sole (dismissed):

src/dpmcore/orm/release_sort_order.py
ruizmaa: Fixed

src/dpmcore/orm/release_sort_order.py
ruizmaa: Fixed, I made it mandatory

src/dpmcore/services/structure.py
ruizmaa: Fixed, guarded both cases explicitly (`limit=None` returns all from offset, negative clamps to 0)

src/dpmcore/orm/release_sort_order.py
ruizmaa: Fixed in [b1e93bc516d18ca0dc7c6a35c140a0634f0755ef](https://github.com/Meaningful-Data/dpmcore/pull/296/changes/b1e93bc516d18ca0dc7c6a35c140a0634f0755ef)

src/dpmcore/dpm_xl/utils/filters.py
ruizmaa: Fixed,moved the session check to the top of `filter_by_release` so it applies directly to the `active_only_fallback=True` path too

andres-sole (approved):

PR #303, opened 2026-08-12
feat(semantic): expose operations_data and validate precondition links

### LINKED ISSUE #302: `SemanticService` gaps around `operations_data` and precondition-operation validation
### Initial Checks

### Problem Description

Two related gaps in `SemanticService`, needed by `drr_operations`

**1. `operations_data` isn't published.** `SemanticService.validate()` publishes `ast`, `oc_data`, `oc_tables` and `oc_parameters` as per-call state. The underlying `OperandsChecking` pass also computes `operations_data` (a DataFrame from `OperationQuery.get_operations_from_codes`, populated whenever the expression contains a `TemporaryAssignment`), but `SemanticService` never surfaces it, even though `MLGeneration` already accepts it as a constructor argument alongside `data` and uses it (`_get_op_version_id`).

**2. No way to validate an externally-linked precondition operation.** `OperationVersion.precondition_operation_vid` (a self-FK to a separate, already-persisted precondition operation) is never read for validation, only written (`ecb_validations_import.py`) and serialised (`structure.py`). Validating it against the current expression needs two checks:

* every filing-indicator variable the linked operation's persisted expression tree references must still be a parameter of some currently-open module;
* every precondition item embedded in the linked operation's own expression text must map (via its abstract-table code) to one of the current expression's tables, and its modules must be a subset of the current expression's tables' modules.

### Proposed Solution

```python
1. Expose `oc_operations_data` on `SemanticService`, mirroring `oc_data`/`oc_tables`/`oc_parameters`.
2. Add a keyword-only `precondition_operation_vid` to `validate()`; when given and the expression is otherwise valid, run the two checks above (new error codes `7-3`/`7-4`/`7-5`). A VID with no matching `OperationVersion` is accepted with nothing to check.
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #302

## Summary

Adds `oc_operations_data` to `SemanticService` and a new `precondition_operation_vid` argument to `validate()`, with the two checks it needs (`7-3`/`7-4`/`7-5`).

- `services/semantic.py`: publish `oc_operations_data` (mirrors `oc_data`/`oc_tables`/`oc_parameters`). Added `precondition_operation_vid` param and `_check_precondition_link`
- `dpm_xl/model_queries.py`: new `TableVersionQuery.get_abstract_table_codes`.
- `errors.py`: new `7-3`/`7-4`/`7-5` messages.

## Checklist

- [] Documentation updated (if applicable)

## Impact / Risk

### DIFF
--- a/src/dpmcore/services/semantic.py
+++ b/src/dpmcore/services/semantic.py
@@ -750,10 +750,21 @@ class SemanticService:
         table_ok: Callable[[str], bool],
         release_id: Optional[int],
     ) -> None:
-        """Raise ``7-5`` when the gate's modules don't fit (``7-4`` passed)."""
+        """Raise ``7-5`` when the gate's modules don't fit (``7-4`` passed).
+
+        A gate code matched at the abstract level is expanded to its
+        concrete children first, since module composition never links
+        an abstract table directly.
+        """
+        concrete_by_code = TableVersionQuery.get_concrete_table_codes(
+            self.session, list(precondition_tables), release_id
+        )
+        concrete_precondition_tables: set[str] = set().union(
+            *concrete_by_code.values()
+        )
         precondition_modules_df = ModuleVersionQuery.get_from_table_codes(
             self.session,
-            list(precondition_tables),
+            list(concrete_precondition_tables),
             release_id,
             include_ghosts=True,
         )
@@ -768,14 +779,23 @@ class SemanticService:
             if not operand_modules_df.empty
             else set()
         )
-        precondition_modules_by_table: dict[str, set[str]] = {}
+        modules_by_concrete_table: dict[str, set[str]] = {}
         if not precondition_modules_df.empty:
             for table_code, group in precondition_modules_df.groupby(
                 "TableCode"
             ):
-                precondition_modules_by_table[str(table_code)] = set(
+                modules_by_concrete_table[str(table_code)] = set(
                     group["ModuleCode"]
                 )
+        precondition_modules_by_table: dict[str, set[str]] = {
+            code: set().union(
+                *(
+                    modules_by_concrete_table.get(concrete, set())
+                    for concrete in concretes
+                )
+            )
+            for code, concretes in concrete_by_code.items()
+        }
 
         def module_ok(code: str) -> bool:
             if code not in precondition_abstract_by_table:
--- a/src/dpmcore/services/semantic.py
+++ b/src/dpmcore/services/semantic.py
@@ -3,7 +3,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field, replace
-from typing import TYPE_CHECKING, Any, Optional, cast
+from typing import TYPE_CHECKING, Any, Callable, Optional, cast
 
 from sqlalchemy import func
 
@@ -30,7 +30,10 @@ from dpmcore.orm.operations import (
 from dpmcore.orm.query_utils import chunked_in
 from dpmcore.orm.variables import VariableVersion
 from dpmcore.services._parameters import merge_parameters
-from dpmcore.services._precondition_codes import required_precondition_codes
+from dpmcore.services._precondition_codes import (
+    extract_precondition_codes,
+    gate_satisfiable,
+)
 from dpmcore.services.syntax import SyntaxService
 
 if TYPE_CHECKING:
@@ -320,9 +323,9 @@ class SemanticService:
                 )
             except SemanticError as exc:
                 code = getattr(exc, "code", None)
-                return self._failure(expression, exc, code)
+                return self._precondition_link_failure(result, exc, code)
             except Exception as exc:
-                return self._failure(expression, exc, "UNKNOWN")
+                return self._precondition_link_failure(result, exc, "UNKNOWN")
         return result
 
     # ------------------------------------------------------------------ #
@@ -425,28 +428,53 @@ class SemanticService:
         except Exception as exc:
             return self._failure(expression, exc, "UNKNOWN")
 
-    def _failure(
-        self,
-        expression: str,
-        exc: Exception,
-        error_code: Optional[str],
-    ) -> SemanticResult:
-        """Clear the published per-call state and build a failing result.
+    def _clear_published_state(self) -> None:
+        """Clear per-call state exposed on ``self`` after a failure.
 
-        ``ast`` is cleared alongside ``oc_*``; leaving the previous call's AST
-        readable after a failure invited consumers to act on stale state.
+        Leaving the previous call's AST/``oc_*`` readable after a failure
+        invited consumers to act on stale state.
         """
         self.ast = None
         self.oc_data = None
         self.oc_tables = None
         self.oc_parameters = None
         self.oc_operations_data = None
+
+    def _failure(
+        self,
+        expression: str,
+        exc: Exception,
+        error_code: Optional[str],
+    ) -> SemanticResult:
+        """Clear the published per-call state and build a failing result."""
+        self._clear_published_state()
         return SemanticResult(
             is_valid=False,
             error_message=str(exc),
             error_code=error_code,
             expression=expression,
             error_source="expression",
+        )
+
+    def _precondition_link_failure(
+        self,
+        result: SemanticResult,
+        exc: Exception,
+        error_code: Optional[str],
+    ) -> SemanticResult:
+        """Fail result over a precondition-link mismatch.
+
+        Unlike :meth:`_failure`, ``expression``/gate already validated, so
+        the existing halves survive and ``error_source`` is
+        ``"precondition"`` rather than ``"expression"``.
+        """
+        self._clear_published_state()
+        return replace(
+            result,
+            is_valid=False,
+            error_message=str(exc),
+            error_code=error_code,
+            error_source="precondition",
         )
 
     def _resolution_failure(
@@ -577,8 +605,8 @@ class SemanticService:
         and its tables/modules fit the current expression's (``7-4``/
         ``7-5``). Nothing to check against (malformed, closed, or an
         unresolved ``precondition_operation_vid``) is skipped, not
-        raised. Only codes ``required_precondition_codes`` deems
-        mandatory count, so ``{v_A} or {v_B}`` needs just one side.
+        raised. An ``or``/``xor`` only fails a check when every side
+        does, so ``{v_A} or {v_B}`` needs just one side to pass.
         """
         row = filter_by_release(
             self.session.query(OperationVersion.expression).filter(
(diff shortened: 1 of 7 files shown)


src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,138 @@ def is_valid(
...
+                VariableVersion.end_release_id,
+                release_id,
+            ).all()
+        }
+        filing_indicator_codes = ModuleVersionQuery.get_filing_indicator_codes(
+            self.session, codes
+        )
+        if not filing_indicator_codes:
+            return
+        modules_df = ModuleVersionQuery.get_precondition_module_versions(
```
andres-sole, line 650: This methods was intended to calcuate scopes, so it omits mvs with FromReferenceDate == ToReferenceDate (ghosts). So it can give us the answer that there a no modules when there are, but they are "ghosts". I think we should discuss this. @ruizmaa

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,138 @@ def is_valid(
...
+        )
+        operand_abstract_tables = set(abstract_by_table.values())
+        if not set(precondition_items).issubset(operand_abstract_tables):
+            raise SemanticError(
+                "7-4",
+                precondition_tables=", ".join(precondition_items),
+                operation_tables=", ".join(sorted(operand_abstract_tables)),
+            )
+
+        precondition_modules_df = ModuleVersionQuery.get_from_table_codes(
```
andres-sole, line 765: Same mechanism (omits ghost mvs)

andres-sole (commented):
I think this should be discussed, i think for module versions membership the check should include ghosts modules, even if the resulting scopes are not valid

src/dpmcore/services/semantic.py
ruizmaa: I think ghosts should count differently here than in scope calculation. What if `get_from_table_codes`/`get_precondition_module_versions` took an include_ghosts flag that skips the fallback and returns the ghost itself, defaulting to `False` so scope calculation stays untouched? Thoughts?

andres-sole: It sounds great, but my concern is that the API is not clear enough. If we are going this route docs should be clear about what does what

ruizmaa: Addressed it in the docstrings on `_resolve_covering`, `get_from_table_codes`, and `get_precondition_module_versions`. Let me know if that's clear enough.

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+
+        A malformed expression is skipped, not raised. Uses
+        ``extract_precondition_codes`` since dpmcore's parser produces
+        ``VarRef``, not ``PreconditionItem``, for these references.
+        """
+        try:
+            precondition_ast = self._syntax.parse(precondition_expression)
+        except Exception:
+            return
+        precondition_items = extract_precondition_codes(precondition_ast)
```
andres-sole: `_check_precondition_tables` uses all codes of the gate, without keeping only the filing indicators. A value condition like `{v_BM} = 'G-SIB'` gives the code `BM`. `BM` is not a table, so 7-4 is raised every time

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+        if not precondition_items:
+            return
+
+        operand_tables = list(self.oc_tables.keys()) if self.oc_tables else []
+        abstract_by_table = TableVersionQuery.get_abstract_table_codes(
+            self.session, operand_tables, release_id
+        )
+        operand_abstract_tables = set(abstract_by_table.values())
+        if not set(precondition_items).issubset(operand_abstract_tables):
```
andres-sole: `{v_C_01.00} or {v_C_02.00}` on a module that only has `C_01.00` fails, because `issubset` asks for both codes. Same problem at `:642` for 7-3. `_precondition_codes.py` already solves this

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+            return
+        precondition_items = extract_precondition_codes(precondition_ast)
+        if not precondition_items:
+            return
+
+        operand_tables = list(self.oc_tables.keys()) if self.oc_tables else []
+        abstract_by_table = TableVersionQuery.get_abstract_table_codes(
+            self.session, operand_tables, release_id
+        )
+        operand_abstract_tables = set(abstract_by_table.values())
```
andres-sole, line 712: `operand_abstract_tables = set(abstract_by_table.values())` keeps only the abstract parents

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+        """
+        try:
+            precondition_ast = self._syntax.parse(precondition_expression)
+        except Exception:
+            return
+        precondition_items = extract_precondition_codes(precondition_ast)
+        if not precondition_items:
+            return
+
+        operand_tables = list(self.oc_tables.keys()) if self.oc_tables else []
```
andres-sole, line 706: an expression without tables always fails 7-4

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+                    OperationNode.operation_vid == precondition_operation_vid,
+                    OperandReference.variable_id.isnot(None),
+                )
+                .distinct()
+                .all()
+            }
+        )
+        if not variable_ids:
+            return
+        codes = {
```
andres-sole: The variable codes are filtered by release before we know if they are filing indicators. A retired variable gives no code

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+        )
+        operand_abstract_tables = set(abstract_by_table.values())
+        if not set(precondition_items).issubset(operand_abstract_tables):
+            raise SemanticError(
+                "7-4",
+                precondition_tables=", ".join(precondition_items),
+                operation_tables=", ".join(sorted(operand_abstract_tables)),
+            )
+
+        precondition_modules_df = ModuleVersionQuery.get_from_table_codes(
```
andres-sole, line 765: After 7-4, the gate codes are abstract table codes. But `get_from_table_codes` only finds concrete tables

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,147 @@ def is_valid(
...
+        """Cross-check an externally-linked precondition operation.
+
+        Its persisted filing-indicator variables must still be live
+        (``7-3``), and its own expression's tables/modules must fit the
+        current expression's (``7-4``/``7-5``).
+        """
+        self._check_precondition_filing_indicators(
+            precondition_operation_vid, release_id
+        )
+        expression_row = (
```
andres-sole: There is no check that `precondition_operation_vid` exists, and no release filter

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,167 @@ def is_valid(
...
+            OperationVersion.end_release_id,
+            release_id,
+        ).first()
+        if row is None or not row[0]:
+            return
+        try:
+            precondition_ast = self._syntax.parse(row[0])
+        except Exception:
+            return
+        required_codes = set(required_precondition_codes(precondition_ast))
```
andres-sole: `required_precondition_codes` returns `set()` for a pure `or`, so the whole link check is skipped. Fine for scope, wrong for a gate because an `or` should require at least one live side

src/dpmcore/services/semantic.py
```diff
@@ -286,17 +300,30 @@ def validate(
...
+            result = self._combine(main, self._cross_check(main, precondition))
+
+        if result.is_valid and precondition_operation_vid is not None:
+            try:
+                self._check_precondition_link(
+                    precondition_operation_vid, resolved
+                )
+            except SemanticError as exc:
+                code = getattr(exc, "code", None)
+                return self._failure(expression, exc, code)
```
andres-sole: a failed link check calls `_failure()`, so `precondition` becomes `None`, both warnings are lost and `error_source` says `"expression"`. Use `replace(result, is_valid=False, …, error_source="precondition")`

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,167 @@ def is_valid(
...
+                    )
+                    .join(
+                        VariableVersion,
+                        OperandReference.variable_id
+                        == VariableVersion.variable_id,
+                    )
+                    .filter(
+                        OperationNode.operation_vid
+                        == precondition_operation_vid,
+                        VariableVersion.code.in_(filing_indicator_codes),
```
andres-sole: `code.in_(filing_indicator_codes)` should be `.in_(filing_indicator_codes - live_codes)`

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +564,167 @@ def is_valid(
...
+            return
+        modules_df = ModuleVersionQuery.get_precondition_module_versions(
+            self.session,
+            list(filing_indicator_codes),
+            release_id,
+            include_ghosts=True,
+        )
+        live_codes = set(modules_df["Code"]) if not modules_df.empty else set()
+        if not filing_indicator_codes.issubset(live_codes):
+            variable_ids = sorted(
```
andres-sole: the message is rebuilt from the `OperationNode` join instead of the stale codes already computed, so an operation with no node rows raises 7-3 with `Precondition variable ID: .`

src/dpmcore/services/semantic.py
ruizmaa: Fixed in [`9fbcef3`](https://github.com/Meaningful-Data/dpmcore/pull/303/changes/9fbcef326880882d79ca2c2269b6d5a0517785e1). Dropped `required_precondition_codes` from this path, gating is now on `extract_precondition_codes` and the actual or/xor logic is handled by `gate_satisfiable`. So a pure `{v_A} or {v_B}` is evaluated instead of skipped

src/dpmcore/services/semantic.py
ruizmaa: Fixed in [`9fbcef3`](https://github.com/Meaningful-Data/dpmcore/pull/303/changes/9fbcef326880882d79ca2c2269b6d5a0517785e1). New `_precondition_link_failure` uses `replace(result, is_valid=False, ..., error_source="precondition")` instead of `_failure()`

src/dpmcore/services/semantic.py
ruizmaa: Fixed in [`9fbcef3`](https://github.com/Meaningful-Data/dpmcore/pull/303/changes/9fbcef326880882d79ca2c2269b6d5a0517785e1). Query now filters on `stale_codes` (`filing_indicator_codes - live_codes`)

src/dpmcore/services/semantic.py
ruizmaa: Fixed in [`9fbcef3`](https://github.com/Meaningful-Data/dpmcore/pull/303/changes/9fbcef326880882d79ca2c2269b6d5a0517785e1). Falls back to the `stale_codes` when the join returns no variable ids

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +592,210 @@ def is_valid(
...
+        precondition_modules_by_table: dict[str, set[str]] = {}
+        if not precondition_modules_df.empty:
+            for table_code, group in precondition_modules_df.groupby(
+                "TableCode"
+            ):
+                precondition_modules_by_table[str(table_code)] = set(
+                    group["ModuleCode"]
+                )
+
+        def module_ok(code: str) -> bool:
```
andres-sole, line 800: 7-5 undoes what 7-4 does. Resolve the modules from the concrete tables that mapped to the abstract one, not from the abstract code

src/dpmcore/services/semantic.py
```diff
@@ -535,6 +592,210 @@ def is_valid(
...
+            if code not in precondition_abstract_by_table:
+                return True
+            modules = precondition_modules_by_table.get(code, set())
+            return bool(modules & operand_modules)
+
+        def fits(code: str) -> bool:
+            return table_ok(code) and module_ok(code)
+
+        if not gate_satisfiable(precondition_ast, fits):
```
andres-sole, line 809: module_ok needs a guard like the one on (`if not precondition_abstract_by_table: return`). `

src/dpmcore/services/semantic.py
ruizmaa: Added `TableVersionQuery.get_concrete_table_codes` to expand each precondition table to its concrete children before resolving modules

src/dpmcore/services/semantic.py
ruizmaa: Added `if precondition_modules_df.empty: return` at the top of `_check_precondition_modules`

andres-sole (approved):

PR #308, opened 2026-08-14
Fix/release filter numeric comparison

### LINKED ISSUE #307: Release-filtered queries compare ReleaseID as a plain number instead of by date
### Initial Checks

### Summary

Several `model_queries.py` query helpers filter by `release_id` with a raw numeric comparison (`start_release_id <= release_id`) instead of the date-ordered `filter_by_release` helper `ModuleVersionQuery` already uses.
`ReleaseID` isn't chronological, so this drops any row whose `start_release_id` is numerically larger than `release_id` even when it's chronologically earlier. Real variables, items, operations and table versions get reported as not found.

### Reproducible Example

```python
from datetime import date

from dpmcore.orm import Base
from dpmcore.orm.variables import Variable, VariableVersion
from dpmcore.orm.glossary import ItemCategory
from dpmcore.orm.rendering import Table, TableVersion
from dpmcore.dpm_xl.model_queries import (
    VariableVersionQuery,
    ItemCategoryQuery,
    TableVersionQuery,
)

    session.add_all([
        Release(release_id=1, code="4.2", date=date(2025, 10, 31)),
        # Post-4.2.1 releases use a non-monotonic id scheme
        # (real 4.2.1 = ReleaseID 1010000003).
        Release(release_id=1010000050, code="4.3", date=date(2026, 6, 28)),
        # Legacy perpetual-open sentinel: small numeric id, but its type
        # makes it sort as "latest" regardless of its (placeholder) date.
        Release(release_id=9999, code="Playground", date=date(1970, 1, 1), type="playground"),

        # Each row below was introduced in 4.3 -- StartReleaseID is
        # numerically larger than 9999, though chronologically it is
        # before it. Same bug, three independent query classes, three
        # different tables.
        Variable(variable_id=1),
        VariableVersion(variable_id=1, variable_vid=1, code="W_04.10",
                         start_release_id=1010000050, end_release_id=None),

        ItemCategory(item_id=1, signature="eba_qEH:qx2005", code="qx2005",
                     category_id=1, start_release_id=1010000050, end_release_id=None),

        Table(table_id=1),
        TableVersion(table_vid=1, table_id=1, code="F_20.04",
                     start_release_id=1010000050, end_release_id=None),
    ])
    session.commit()

    print(VariableVersionQuery.check_variable_exists(session, "W_04.10", 9999))
    print(ItemCategoryQuery.get_items(session, ["eba_qEH:qx2005"], 9999).empty)
    print(TableVersionQuery.check_table_exists(session, "F_20.04", 9999))
```

### Expected vs Actual

Expected: `True`, `False` (not empty), `True`. Each row is open since release `4.3`, which chronologically precedes the perpetual `Playground` release, so it must still exist when queried at `Playground` too.

Actual: all three fail identically. `check_variable_exists` prints `False`, `get_items` returns an empty DataFrame, `check_table_exists` prints `False`. Same root cause (`start_release_id <= release_id` comparing `1010000050 <= 9999`, numerically false despite `4.3` being chronologically earlier), three independent query classes, three different tables.

### dpmcore version

0.1.2rc8

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #307 

## Summary

Replaces the raw `start_release_id <= release_id` comparison with the date-ordered `filter_by_release` helper (matching `ModuleVersionQuery`) in `VariableVersionQuery.check_variable_exists`/`check_precondition`/`get_variable_id`/`get_all_preconditions`, `ItemCategoryQuery.get_items` and `TableVersionQuery.check_table_exists`. The call sites reachable from `SemanticService.validate()` and scope calculation. `release_id=None` behavior is untouched everywhere.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -228,14 +228,11 @@ class ItemCategoryQuery:
         if items:
             query = query.filter(ItemCategory.signature.in_(items))
         if release_id is not None:
-            query = query.filter(
-                and_(
-                    ItemCategory.start_release_id <= release_id,
-                    or_(
-                        ItemCategory.end_release_id > release_id,
-                        ItemCategory.end_release_id.is_(None),
-                    ),
-                )
+            query = filter_by_release(
+                query,
+                start_col=ItemCategory.start_release_id,
+                end_col=ItemCategory.end_release_id,
+                release_id=release_id,
             )
         else:
             query = query.filter(ItemCategory.end_release_id.is_(None))
@@ -344,14 +341,11 @@ class VariableVersionQuery:
             VariableVersion.code == variable_code
         )
         if release_id is not None:
-            query = query.filter(
-                and_(
-                    VariableVersion.start_release_id <= release_id,
-                    or_(
-                        VariableVersion.end_release_id > release_id,
-                        VariableVersion.end_release_id.is_(None),
-                    ),
-                )
+            query = filter_by_release(
+                query,
+                start_col=VariableVersion.start_release_id,
+                end_col=VariableVersion.end_release_id,
+                release_id=release_id,
             )
         else:
             query = query.filter(VariableVersion.end_release_id.is_(None))
@@ -392,16 +386,12 @@ class VariableVersionQuery:
                 _is_filing_indicator(),
             )
         )
-        if release_id is not None:
-            query = query.filter(
-                and_(
-                    VariableVersion.start_release_id <= release_id,
-                    or_(
-                        VariableVersion.end_release_id > release_id,
-                        VariableVersion.end_release_id.is_(None),
-                    ),
-                )
-            )
+        query = filter_by_release(
+            query,
+            start_col=VariableVersion.start_release_id,
+            end_col=VariableVersion.end_release_id,
+            release_id=release_id,
+        )
         return query.first()
 
     @staticmethod
@@ -423,16 +413,12 @@ class VariableVersionQuery:
         query = session.query(VariableVersion.variable_id).filter(
             VariableVersion.code == value
         )
-        if release_id is not None:
-            query = query.filter(
-                and_(
-                    VariableVersion.start_release_id <= release_id,
-                    or_(
-                        VariableVersion.end_release_id > release_id,
-                        VariableVersion.end_release_id.is_(None),
-                    ),
-                )
-            )
+        query = filter_by_release(
+            query,
+            start_col=VariableVersion.start_release_id,
+            end_col=VariableVersion.end_release_id,
+            release_id=release_id,
+        )
         rows = query.all()
         if not rows:
             return None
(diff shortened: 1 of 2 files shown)


src/dpmcore/dpm_xl/model_queries.py
andres-sole: THis still uses start_release_id <= release_id

andres-sole: Same comparison

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -610,16 +592,12 @@ def check_table_exists(
...
-            query = query.filter(
-                and_(
-                    TableVersion.start_release_id <= release_id,
-                    or_(
-                        TableVersion.end_release_id > release_id,
-                        TableVersion.end_release_id.is_(None),
-                    ),
-                )
-            )
+        query = filter_by_release(
```
andres-sole, line 589: dropping the `if release_id is not None:` guard makes an unknown release raise a bare `ValueError`. We either safeguard it before or add an specific exception

src/dpmcore/dpm_xl/model_queries.py
ruizmaa: Fixed in [`b90d004`](https://github.com/Meaningful-Data/dpmcore/pull/308/commits/b90d00498300f4caac74ba710246b3eb40823cdd), now uses `filter_by_release`.

src/dpmcore/dpm_xl/model_queries.py
ruizmaa: Fixed in [`9b2c3c4`](https://github.com/Meaningful-Data/dpmcore/pull/308/commits/9b2c3c49a25f75ddbb17d3f52208a79bcf4c8501), caught the `ValueError`, returns `False` for an unknown `release_id` (same behavior as the old raw comparison)

src/dpmcore/dpm_xl/model_queries.py
ruizmaa: Same fix, in [`3a94ef2`](https://github.com/Meaningful-Data/dpmcore/pull/308/commits/3a94ef2b4d177ce51eb029a8587b432d729e2bfe)

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -605,21 +581,20 @@ def check_table_exists(
...
-                )
+        try:
+            query = filter_by_release(
+                query,
+                start_col=TableVersion.start_release_id,
+                end_col=TableVersion.end_release_id,
+                release_id=release_id,
             )
+        except ValueError:
+            return False
```
andres-sole: the other  filter_by_release calls propagate the exception. I think they should be uniform

tests/integration/releases/test_release_sort_order.py
```diff
@@ -915,3 +917,148 @@ def test_resolve_explicit_release_self_reference_at_playground_type(
...
+
+
+# --------------------------------------------------------------------- #
+# VariableVersionQuery / ItemCategoryQuery / TableVersionQuery
+# release filters must date-order, not numerically compare, release_id
+# --------------------------------------------------------------------- #
+
+
+@pytest.fixture
+def non_monotonic_id_session(memory_session):
```
andres-sole, line 929: Fixture has only rows that must match. So we are not testing the release filtering in the new tests

src/dpmcore/dpm_xl/model_queries.py
ruizmaa: I agree, just fixed

tests/integration/releases/test_release_sort_order.py
ruizmaa: Fixed, the fixture now includes a row that must be excluded (dated earlier despite a higher release_id), with assertions across all six helpers

andres-sole (approved):

PR #315, opened 2026-08-19
feat(dpm_xl): serialize set operators under a single SetOp class name

### LINKED ISSUE #314: Collapse the five set operators into a single `SetOp` class name in serialized scripts
### Summary

Serialized scripts give each set operator its own `class_name`, mirroring the internal AST class, and each one carries its operands under a different field:

| DPM-XL | current `class_name` | `op` | current operand field(s) |
| --- | --- | --- | --- |
| `set_of(x)` | `SetOfOp` | `set_of` | `operand` |
| `union(a, b, …)` | `UnionSetOp` | `union` | `operands` |
| `intersect(a, b, …)` | `IntersectSetOp` | `intersect` | `operands` |
| `setdiff(a, b)` | `SetdiffOp` | `setdiff` | `left` + `right` |
| `symdiff(a, b)` | `SymdiffOp` | `symdiff` | `left` + `right` |

The per-operator name only replicates what `op` already says, and `class_name` is expected to drop out of the script contract eventually, so the engine side prefers one shared name for the whole family with `op` as the discriminator. That also brings the family in line with the rest of the serializer: `BinOp`, `UnaryOp`, `AggregationOp` and `ComplexNumericOp` are all generic class names discriminated by `op`, and the set operators are the only outlier.

Collapse all five onto `SetOp`, and unify the operand fields into a single `operands` array so the family has one shape rather than three:

```json
{"class_name": "SetOp", "op": "set_of",    "operands": [{...}]}
{"class_name": "SetOp", "op": "union",     "operands": [{...}, {...}, {...}]}
{"class_name": "SetOp", "op": "intersect", "operands": [{...}, {...}]}
{"class_name": "SetOp", "op": "setdiff",   "operands": [{...}, {...}]}
{"class_name": "SetOp", "op": "symdiff",   "operands": [{...}, {...}]}
```

`setdiff` and `symdiff` become positional: `operands[0]` is the former `left`, `operands[1]` the former `right`. Order is significant for `setdiff`, so it has to be preserved on emission and asserted in the tests.

The array length does not need validating downstream — the grammar already fixes the arity of each operator and rejects anything else at parse time, so it can never reach a script:

- `set_of` — exactly one operand (`set_of(a, b)` is a syntax error)
- `union` / `intersect` — two or more (`union(a)` is a syntax error)
- `setdiff` / `symdiff` — exactly two (`setdiff(a, b, c)` is a syntax error)

Set literals are unchanged: `{"class_name": "Set", "children": [...]}`. `Set` is a literal, not an operator.

### Scope

- `src/dpmcore/dpm_xl/utils/serialization.py` — the five `ASTToJSONVisitor` visitors collapse into one emission shape. This is the only place these names are produced: every AST node routes through this visitor, so there is no second path to keep in sync.
- Internal AST class names (`SetOfOp`, `UnionSetOp`, `IntersectSetOp`, `SetdiffOp`, `SymdiffOp`) and their `left`/`right`/`operand`/`operands` attributes stay as they are. They are dpmcore-internal, and renaming them would ripple into the constructor, semantic analyzer, template and ML generation for no gain on the wire.

Out of scope:

- The operation-node/database path (`MLGeneration`) resolves operators by symbol rather than by class name, so it needs no change.
- `count(set)` already serializes as `{"class_name": "AggregationOp", "op": "count"}`. The legacy `CountSetOp` AST node is not emitted by script generation.
- `deserialize_ast` resolves `class_name` by lookup against the AST module, so `SetOp` will not map back to a node — the same as the already-emitted `GetClauseOp`, `SubClauseOp`, `RenameClauseOp`, `ConditionalOp` and `Function`. Nothing in the library or the suite calls it. If we want JSON round-tripping to work it needs `op`-based dispatch for every collapsed family, which is its own issue rather than a partial fix here.

### Testing

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Serialized scripts gave each set operator its own `class_name` mirroring the internal AST class — `SetOfOp`, `UnionSetOp`, `IntersectSetOp`, `SetdiffOp`, `SymdiffOp` — and carried the operands under three different fields: `operand` for `set_of`, `operands` for `union`/`intersect`, `left` plus `right` for `setdiff`/`symdiff`. The per-operator name only repeated what `op` already said, and it left the family as the only outlier in a serializer where `BinOp`, `UnaryOp`, `AggregationOp` and `ComplexNumericOp` are all generic class names discriminated by `op`.

All five now serialize as `SetOp` with a single positional `operands` array:

```json
{"class_name": "SetOp", "op": "set_of",    "operands": [{...}]}
{"class_name": "SetOp", "op": "union",     "operands": [{...}, {...}, {...}]}
{"class_name": "SetOp", "op": "intersect", "operands": [{...}, {...}]}
{"class_name": "SetOp", "op": "setdiff",   "operands": [{...}, {...}]}
{"class_name": "SetOp", "op": "symdiff",   "operands": [{...}, {...}]}
```

`setdiff` and `symdiff` become positional: `operands[0]` is the former `left`, `operands[1]` the former `right`. Order is significant for `setdiff`, so it is preserved on emission and asserted with asymmetric operands in the tests. Set literals are untouched — `{"class_name": "Set", "children": [...]}` — and `count(set)` still serializes as `{"class_name": "AggregationOp", "op": "count"}`.

The array length needs no validating downstream: the grammar fixes the arity of each operator and rejects anything else at parse time, so `set_of(a, b)`, `union(a)` and `setdiff(a, b, c)` can never reach a script.

Internal AST class names and their `left`/`right`/`operand`/`operands` attributes stay as they are. They are dpmcore-internal, and renaming them would ripple into the constructor, semantic analyzer, template and ML generation for no gain on the wire. The operation-node path resolves operators by symbol rather than by class name, so it needs no change.

Fixes #314.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/utils/serialization.py
+++ b/src/dpmcore/dpm_xl/utils/serialization.py
@@ -625,30 +625,17 @@ class ASTToJSONVisitor(NodeVisitor):
     visit_IntersectSetOp = visit_UnionSetOp
     visit_SymdiffOp = visit_SetdiffOp
 
-    def visit_CountSetOp(self, node: Any) -> NodeDict:
-        """Visit legacy ``CountSetOp`` nodes as a ``count`` aggregation.
-
-        ``count`` is not a set operator — it returns the set's
-        cardinality as a Scalar — and its dedicated grammar rule was
-        dropped in MR !74, so ``count(...)`` now parses to an
-        ``AggregationOp``. The class survives only for externally built
-        ASTs, so emit the ``AggregationOp`` shape the parser would have
-        produced rather than let ``generic_visit`` invent a
-        ``CountSetOp`` class name the consumer schema rejects.
-
-        Args:
-            node: The legacy ``CountSetOp`` AST node.
-
-        Returns:
-            dict: The serialized ``AggregationOp`` node.
-        """
-        return {
-            "class_name": "AggregationOp",
-            "op": node.op,
-            "operand": self.visit(node.operand),
-            "grouping_clause": None,
-            "analytic_clause": None,
-        }
+    # ``count`` is not a set operator — it returns the set's cardinality
+    # as a Scalar — and its dedicated grammar rule was dropped in MR !74,
+    # so ``count(...)`` now parses to an ``AggregationOp``. The legacy
+    # class survives only for externally built ASTs, so alias it to the
+    # ``AggregationOp`` handler: it carries ``op``/``operand`` and no
+    # grouping or analytic clause, which is exactly what that handler's
+    # ``hasattr`` checks already emit (both clauses null). Aliasing keeps
+    # the parser shape and the legacy shape from drifting apart, and
+    # stops ``generic_visit`` inventing a ``CountSetOp`` class name the
+    # consumer schema rejects.
+    visit_CountSetOp = visit_AggregationOp
 
     def visit_ParExpr(self, node: Any) -> NodeDict:
         """Visit ParExpr nodes."""
--- a/src/dpmcore/dpm_xl/utils/serialization.py
+++ b/src/dpmcore/dpm_xl/utils/serialization.py
@@ -584,47 +584,48 @@ class ASTToJSONVisitor(NodeVisitor):
             "children": [self.visit(child) for child in node.children],
         }
 
+    def _visit_set_op(self, node: Any, operands: list[Any]) -> NodeDict:
+        """Emit a set operator under the shared ``SetOp`` class name.
+
+        The whole family shares one class name discriminated by ``op``,
+        with the operands in a single positional array: ``set_of``
+        carries one, ``union``/``intersect`` two or more, and
+        ``setdiff``/``symdiff`` exactly two in left-then-right order
+        (significant for ``setdiff``). The arity is fixed by the
+        grammar, so it needs no validating downstream.
+
+        Args:
+            node: The set operator AST node.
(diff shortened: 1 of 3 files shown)


src/dpmcore/dpm_xl/utils/serialization.py
```diff
@@ -584,46 +584,70 @@ def visit_Set(self, node: Any) -> NodeDict:
...
-        """Visit SymdiffOp nodes (``symdiff(left, right)``)."""
-        return {
-            "class_name": "SymdiffOp",
+            "class_name": "AggregationOp",
             "op": node.op,
-            "left": self.visit(node.left),
-            "right": self.visit(node.right),
+            "operand": self.visit(node.operand),
+            "grouping_clause": None,
+            "analytic_clause": None,
```
ruizmaa: `visit_CountSetOp` duplicates the shape of `visit_AggregationOp`. Since `CountSetOp` has no grouping_clause/analytic_clause attributes, `visit_AggregationOp`'s existing hasattr(...) checks already produce this exact same dict. Could we just use `visit_CountSetOp = visit_AggregationOp`, in line with the alias pattern already used for `visit_IntersectSetOp`/`visit_SymdiffOp`?

andres-sole: Yes thats better, just changed it

ruizmaa (approved):

PR #323, opened 2026-08-20
Prefer the adopted TableVersion when Playground stages a future draft

### LINKED ISSUE #320: `get_last_release`/`filter_by_release` return two ModuleVersions when a future version is staged at the "Playground" release too
### Initial Checks

### Summary

When resolving the current release (`release_id=None`, or explicitly `release_id=<playground release>`), `filter_by_release` can match two `ModuleVersion` rows for the same entity at once: the one genuinely active today, and a future version whose `FromReferenceDate` is still ahead of today and has not yet been assigned a real, numbered release id.

This happens because the "playground" release is used as both the end-marker for the currently active row and the start-marker for the future row. Each use is individually correct and tested on its own; they only collide when both rows reference that same release at once.

### Reproducible Example

```python
from datetime import date

from dpmcore.orm import Base
from dpmcore.orm.packaging import Framework, Module, ModuleVersion
from dpmcore.dpm_xl.model_queries import ModuleVersionQuery
from dpmcore.dpm_xl.utils.filters import filter_by_release

    session.add_all([
        # A real, dated release, plus the DB's "Playground" release, typed
        # so dpmcore always treats it as the latest, regardless of its date.
        Release(release_id=5, code="4.2", date=date(2025, 10, 31)),
        Release(
            release_id=9999,
            code="Playground",
            date=date(1970, 1, 1),
            type="playground",
        ),
        Framework(framework_id=1, code="FW"),
        Module(module_id=1, framework_id=1),
        # Currently active: opened at release 5, closed via the DB's own
        # "not yet tied to a numbered release" sentinel (end=9999).
        ModuleVersion(
            module_vid=1,
            module_id=1,
            code="MV_ACTIVE",
            start_release_id=5,
            end_release_id=9999,
            from_reference_date=date(2026, 3, 31),
        ),
        # A future version, staged ahead of its own activation date, not
        # yet assigned a real release id, so it is parked at 9999 too.
        ModuleVersion(
            module_vid=2,
            module_id=1,
            code="MV_FUTURE",
            start_release_id=9999,
            end_release_id=None,
            from_reference_date=date(2027, 12, 31),
        ),
    ])
    session.commit()

    last_release = ModuleVersionQuery.get_last_release(session)
    print("get_last_release() ->", last_release)

    matched = filter_by_release(
        session.query(ModuleVersion),
        start_col=ModuleVersion.start_release_id,
        end_col=ModuleVersion.end_release_id,
        release_id=last_release,
    ).all()
    print("Matched at the live release:", sorted(mv.code for mv in matched))

    matched_real = filter_by_release(
        session.query(ModuleVersion),
        start_col=ModuleVersion.start_release_id,
        end_col=ModuleVersion.end_release_id,
        release_id=5,
    ).all()
    print("Matched at the real release_id=5:", sorted(mv.code for mv in matched_real))
```

### Expected vs Actual

Expected: at the live release, only `MV_ACTIVE` matches (`MV_FUTURE` isn't active yet, its own reference period hasn't started).
Actual: both `MV_ACTIVE` and `MV_FUTURE` match simultaneously.

### dpmcore version

0.1.2rc8

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #320 

## Summary

At the perpetual ("Playground") release, a table already adopted in a real release and a brand-new draft of the same table can both match `filter_by_release` at once. Merging their cells produced a hybrid structure with false "grey cell" errors.

`ViewDatapointsQuery._resolve_current_table_vids` now prefers the already-adopted `TableVersion`(s) over one just started at the same perpetual release, falling back to every match when none is adopted.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -52,6 +52,7 @@ from dpmcore.orm.packaging import (
 )
 from dpmcore.orm.query_utils import chunked_in
 from dpmcore.orm.release_sort_order import (
+    compute_sort_order,
     load_release_sort_orders,
     release_ids_for_sort_order,
     resolve_sort_order,
@@ -1619,6 +1620,48 @@ class ViewDatapointsQuery:
         }
         return query, aliases
 
+    @staticmethod
+    def _resolve_current_table_vids(
+        session: "Session", table: str, release_id: int | None
+    ) -> list[int]:
+        """Resolve the ``TableVersion.table_vid`` value(s) of ``table`` open at ``release_id``.
+
+        At the perpetual release, an adopted version and one just started
+        there can both compare as "open now". When both are present, only
+        the adopted one(s) are kept.
+
+        Args:
+            session: SQLAlchemy session.
+            table: Table version code.
+            release_id: Release filter; ``None`` resolves to whichever
+                version(s) are currently open.
+
+        Returns:
+            The matching ``table_vid`` values (empty when ``table`` has none
+            open at ``release_id``).
+        """
+        query = session.query(
+            TableVersion.table_vid, TableVersion.start_release_id
+        ).filter(TableVersion.code == table)
+        query = filter_by_release(
+            query,
+            start_col=TableVersion.start_release_id,
+            end_col=TableVersion.end_release_id,
+            release_id=release_id,
+            active_only_fallback=True,
+        )
+        rows = query.all()
+        if len(rows) <= 1:
+            return [row.table_vid for row in rows]
+        sort_orders = load_release_sort_orders(session)
+        perpetual = compute_sort_order(None, None)
+        adopted = [
+            row.table_vid
+            for row in rows
+            if sort_orders.get(row.start_release_id, perpetual) < perpetual
+        ]
+        return adopted or [row.table_vid for row in rows]
+
     @classmethod
     def get_axis_orders(
         cls,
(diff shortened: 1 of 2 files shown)


src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -1889,11 +1922,10 @@ def get_filtered_datapoints(
                     query = query.filter(clause)
 
         if release_id:
-            query = filter_by_release(
```
andres-sole, line 1892: WIthout the filter by release there would be duplicates

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -1805,14 +1846,6 @@ def get_table_data(
                 query, aliases["hvs"].code, sheets, axis_orders["sheets"]
             )
 
-        if release_id is not None:
-            query = filter_by_release(
```
andres-sole, line 1809: without the filter a table which module version is closed at a release now returns data

andres-sole (commented):
the diff also removes the `ModuleVersion` release filter from all three query methods, which the description doesn't mention. I think some clarification is needed in why is so

ruizmaa:
My bad, thanks for flagging it out. I removed the ModuleVersion release filter the three places assuming the new table_vid filter already covered it, but it doesn't.

andres-sole (approved):

PR #327, opened 2026-08-21
fix: declare the shifted home instance for a home-module time_shift

### LINKED ISSUE #325: `time_shift` on a table of the reporting module is classified intra-instance and its reference period is never declared
### Initial Checks

### Summary
When a validation applies `time_shift` to a table that belongs to the module the script is generated for, the generated script declares the validation as intra-instance and declares no dependency at all for the shifted instance. The engine therefore resolves both operands against the current instance, finds no reference period for the shifted one, and the script fails execution.

The shift is detected for the shifted operand, but the reference period built from it is wrong in two independent ways: the sign of the shift is dropped, so `Q, -1` renders as `T-1Q` where the declared period should invert the shift and read `T+1Q`; and the period is only ever attached to a module other than the reporting one, so in this case it is discarded and never reaches the output at all.

COREP_LR 3.1.0 / EGDQ_0896 is the reported case:

```
{tC_48.02, c0020}[get RDT] <= {tC_48.02, c0020}[get refPeriod] and {tC_48.02, c0020}[get RDT] > time_shift({tC_48.02, c0020}, Q, -1, refPeriod)[get refPeriod]
```

`C_48.02` is a COREP_LR table, so the scope resolves to a single module and the validation is reported as intra-instance. The comparison, however, needs a different instance of COREP_LR — `T+1Q`, since the declared period inverts the shift number — and nothing in the script says so.

This is not specific to COREP_LR. Across the currently generated module scripts, seven operations in four scripts are affected:

| module version | operations | shifted table | shift | declared period |
| --- | --- | --- | --- | --- |
| COREP_LR 3.1.0 | EGDQ_0896 | C_48.02 | `Q, -1` | `T+1Q` |
| COREP_OF 3.1.0 | EGDQ_0559, EGDQ_0572 | C_14.00 | `A, 1` | `T-1A` |
| COREP_OF 4.0.0 | EGDQ_0559, EGDQ_0572 | C_14.00 | `A, 1` | `T-1A` |
| COREP_OF 4.1.0 | EGDQ_0559, EGDQ_0572 | C_14.00 | `A, 1` | `T-1A` |

Only the reported case shifts by a negative number, so the inversion changes nothing else in the set. Five of the seven inherit the shifted table from a `with` clause (`with {tC_14.00}: … time_shift({c0120}, A, 1, refPeriod)`), so the shifted operand only resolves to a table after semantic enrichment.

Every `time_shift` in the generated set shifts a table of its own module, so all of them are affected, and `ref_period` is `"T"` in all 74 module entries across every generated script: a non-default reference period has never actually been emitted.

### Reproducible Example

```python
import json

from dpmcore import connect

EXPR = (
    "{tC_48.02, c0020}[get RDT] <= {tC_48.02, c0020}[get refPeriod] and "
    "{tC_48.02, c0020}[get RDT] > "
    "time_shift({tC_48.02, c0020}, Q, -1, refPeriod)[get refPeriod]"
)

    result = db.services.dpm_xl.ast_generator.script(
        expressions=[(EXPR, "EGDQ_0896")],
        module_code="COREP_LR",
        module_version="3.1.0",
    )
    module = list(result["enriched_ast"].values())[0]
    print(json.dumps(module["dependency_information"], indent=2))
    print("dependency_modules:", list(module["dependency_modules"]))
```

### Expected vs Actual
Actual:

```json
{
  "intra_instance_validations": [
    "EGDQ_0896"
  ],
  "cross_instance_dependencies": [],
  "alternative_dependencies": []
}
```

```
dependency_modules: []
```

Expected: the validation is not intra-instance, and the shifted instance of COREP_LR is declared with `ref_period` `T+1Q` so the engine knows which instance to load.

```json
{
  "intra_instance_validations": [],
  "cross_instance_dependencies": [
    {
      "modules": [
        {
          "URI": "http://www.eba.europa.eu/eu/fr/xbrl/crr/fws/corep/its-005-2020/2022-03-01/mod/corep_lr",
          "ref_period": "T+1Q",
          "module_version": "3.1.0"
        }
      ],
      "affected_operations": ["EGDQ_0896"],
      "from_reference_date": "2022-12-31",
      "to_reference_date": "2025-03-30"
    }
  ],
  "alternative_dependencies": []
}
```

`dependency_modules` stays empty: the home module's tables and variables are already the script's top-level blocks, and the shifted instance is the same module version.

### dpmcore version

0.1.2rc9

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Fixes #325.

A validation that applies `time_shift` to a table of the module the script is generated for needs a different instance of that same module. The script said otherwise: the validation was declared intra-instance with no dependency for the shifted operand, so the engine had nothing to resolve it against. The reference period was computed but only ever attached to a module other than the reporting one, which is why `ref_period` was `"T"` in all 74 module entries of the generated scripts.

- A shift on a table the home module owns no longer counts as intra-instance. The shifted instance is declared in `cross_instance_dependencies` under the home module's own URI, with the shift's reference period, module version and reference-date window. A cross-module operation that also shifts a home table gets both entries, the external ones first. No `dependency_modules` entry accompanies the home instance: its tables and variables are the script's top-level blocks already.
- The declared period inverts the shift number: `time_shift(x, A, 1, refPeriod)` needs `T-1A` and `time_shift(x, Q, -1, refPeriod)` needs `T+1Q`. The sign used to be dropped, so a forward shift was inexpressible.
- Cross-instance entries dedupe by `(URI, ref_period)` instead of by URI alone, so two operations needing one module at different instances keep both entries instead of collapsing onto whichever came first.

For COREP_LR 3.1.0 / EGDQ_0896 the block is now the one the issue asks for, with `T+1Q` where the issue wrote `T-1Q`.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/services/ast_generator.py
+++ b/src/dpmcore/services/ast_generator.py
@@ -1345,15 +1345,25 @@ extract_precondition_codes`, shared with
     ) -> None:
         """Merge *new* cross-instance deps into *existing*.
 
-        Deduplicates by the set of module URIs.  When a duplicate is
-        found, its ``affected_operations`` are merged instead.
+        Deduplicates by the set of ``(module URI, reference period)``
+        pairs. When a duplicate is found, its ``affected_operations``
+        are merged instead.
+
+        The reference period is part of the key because two operations
+        can need the *same* module at *different* instances — one at
+        ``T``, another at ``T-1Q`` (#325 makes a non-default period
+        reachable for the home module). Keying on the URI alone merged
+        them into a single entry whose period was whichever operation
+        came first, silently sending the other to the wrong instance.
         """
 
-        def _uri_key(dep: Dict[str, Any]) -> Tuple[str, ...]:
+        def _uri_key(dep: Dict[str, Any]) -> Tuple[Tuple[str, str], ...]:
             modules = dep.get("modules", [])
             return tuple(
                 sorted(
-                    m.get("URI", "") if isinstance(m, dict) else str(m)
+                    (m.get("URI", ""), m.get("ref_period", ""))
+                    if isinstance(m, dict)
+                    else (str(m), "")
                     for m in modules
                 )
             )
@@ -1404,16 +1414,22 @@ extract_precondition_codes`, shared with
 
     @staticmethod
     def _to_ref_period(internal: str) -> str:
-        if internal.startswith("t+"):
+        """Render a ``t(+|-)<indicator><n>`` marker as a reference period.
+
+        The declared period is the *opposite* of the shift the expression
+        asks for. ``time_shift`` moves the operand's reference period by
+        ``shift_number``, so the instance whose shifted operand lines up
+        with the one being reported is the one at ``T - shift_number``:
+        ``time_shift(x, A, 1, refPeriod)`` needs ``T-1A`` and
+        ``time_shift(x, Q, -1, refPeriod)`` needs ``T+1Q``. The marker is
+        built with that inversion already applied (see
+        :meth:`_extract_time_shifts`), so this only reorders it.
+        """
+        if internal.startswith(("t+", "t-")):
+            sign = internal[1]
             ind = internal[2]
             num = internal[3:]
-            if num.startswith("-"):
-                return f"T{num}{ind}"
-            return f"T+{num}{ind}"
-        if internal.startswith("t-"):
-            ind = internal[2]
-            num = internal[3:]
-            return f"T-{num}{ind}"
+            return f"T{sign}{num}{ind}"
         return "T"
 
     @staticmethod
@@ -1438,16 +1454,17 @@ extract_precondition_codes`, shared with
                 prev = current_period[0]
                 pi = node.period_indicator
                 sn = node.shift_number
+                # The declared period inverts the shift number: a shift
+                # of ``+n`` needs the instance ``n`` periods back, a
+                # shift of ``-n`` the one ``n`` periods ahead. A shift
+                # number that is not a literal keeps its direction but
+                # not its size (``n``), which no instance resolves to.
                 if isinstance(sn, Constant):
                     current_period[0] = f"t-{pi}{sn.value}"
                 elif isinstance(sn, UnaryOp) and sn.op == "-":
                     inner = sn.operand
-                    sn_str = (
-                        f"-{inner.value}"
-                        if isinstance(inner, Constant)
-                        else "n"
-                    )
-                    current_period[0] = f"t+{pi}{sn_str}"
+                    num = inner.value if isinstance(inner, Constant) else "n"
+                    current_period[0] = f"t+{pi}{num}"
                 else:
                     current_period[0] = f"t-{pi}n"
                 self.visit(node.operand)
--- a/src/dpmcore/services/ast_generator.py
+++ b/src/dpmcore/services/ast_generator.py
@@ -1433,6 +1433,42 @@ extract_precondition_codes`, shared with
         return "T"
 
     @staticmethod
+    def _literal_shift(node: Any) -> Tuple[int, Optional[int]]:
+        """Resolve a ``shift_number`` expression to ``(sign, magnitude)``.
+
+        ``sign`` is the direction the expression asks for (``1`` or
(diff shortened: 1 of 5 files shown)


src/dpmcore/services/ast_generator.py
ruizmaa, line 1512: `time_shifts` is keyed only by table code, so two `time_shifts` on the same home table with different amounts in one operation silently collide. The second overwrites the first, and one required instance period never gets declared (same failure class as #325)

```python
from dpmcore.dpm_xl.ast.nodes import Constant, VarID, TimeShiftOp, BinOp

shift1 = TimeShiftOp(
    operand=VarID(table="T_SAME", rows=None, cols=None, sheets=None, interval=None, default=None),
    period_indicator="A",
    component=None,
    shift_number=Constant(type_="Integer", value=1),
)
shift2 = TimeShiftOp(
    operand=VarID(table="T_SAME", rows=None, cols=None, sheets=None, interval=None, default=None),
    period_indicator="A",
    component=None,
    shift_number=Constant(type_="Integer", value=2),
)
node = BinOp(left=shift1, op="-", right=shift2)

print(ASTGeneratorService._extract_time_shifts(node))
# -> {'T_SAME': 'T-2A'}
# t-1A is lost
```

Not introduced by this PR, but as is the same issue the PR fixes, could be nice to also fix this

src/dpmcore/services/ast_generator.py
```diff
@@ -1438,16 +1454,17 @@ def visit_TimeShiftOp(self, node: Any) -> None:
...
-                    sn_str = (
-                        f"-{inner.value}"
-                        if isinstance(inner, Constant)
-                        else "n"
-                    )
-                    current_period[0] = f"t+{pi}{sn_str}"
+                    num = inner.value if isinstance(inner, Constant) else "n"
+                    current_period[0] = f"t+{pi}{num}"
                 else:
                     current_period[0] = f"t-{pi}n"
```
ruizmaa: The `shift_number` classification here only handles two shapes, a literal `Constant` (assumed positive) and `UnaryOp(op='-')`, but it doesnt handle the rest:

```python

parse = SyntaxService().parse

node = parse("time_shift({tT1}, Q, +1, refPeriod)").children[0]
print(ASTGeneratorService._extract_time_shifts(node))        # UnaryOp(op='+') -> 'T-nQ' instead of 'T-1Q'

node = parse("time_shift({tT1}, Q, ( -1 ), refPeriod)").children[0]
print(ASTGeneratorService._extract_time_shifts(node))         # ParExpr(UnaryOp('-')) -> 'T-nQ' instead of 'T+1Q'

node = parse("time_shift({tT1}, Q, (-1), refPeriod)").children[0]
print(ASTGeneratorService._extract_time_shifts(node))         # Constant(-1) -> 'T--1Q' (!) instead of 'T+1Q'
```

src/dpmcore/services/scope_calculator.py
```diff
@@ -654,6 +703,106 @@ def detect_cross_module_dependencies(
...
+        tables and variables are already declared at the top level of
+        the script, and the shifted instance is the same module version.
+        """
+        if not periods:
+            return []
+        mv = (
+            self.session.query(ModuleVersion)
+            .filter(ModuleVersion.module_vid == primary_module_vid)
+            .first()
+        )
```
ruizmaa: `_build_home_instance_deps` is called once per operation that shifts a home table, but `primary_module_vid` is constant for the whole script. This re-queries `ModuleVersion` on every call instead of doing it once and reuse it. I guess it would be worth saving it and reusing it

src/dpmcore/services/ast_generator.py
andres-sole: I agree, just fixed it in the latest commit

src/dpmcore/services/ast_generator.py
andres-sole: Fixed

src/dpmcore/services/scope_calculator.py
andres-sole: Fixed in last commit

ruizmaa (approved):

PR #334, opened 2026-08-25
feat: warn on comparisons against items outside a component's domain

### LINKED ISSUE #332: Semantic warning for comparisons against items outside a component's domain
### Initial Checks

### Problem Description

When an enumerated component is compared against an item literal from a different domain, the comparison is always false (`=`, `in`) or always true (`!=`), and the validation silently behaves differently than intended with no diagnostic. Item existence is checked today (error 1-1), but not whether the item belongs to the domain of the component it is compared with.

Nine open operations in EBA DPM 4.2.1 do exactly this:

| Operation | Component → domain | Literal → domain | Effect |
| --- | --- | --- | --- |
| v4968_m, v5024_m, v5372_m, v5373_m, v5376_m | `{tF_00.01, r0010, c0010}` → `qAS` | `eba_AS:x2` → `AS` | condition never true, rule never fires |
| v6169_m | `{tC_101.00, c0070}` → `qIM` | `eba_qPL:qx2078` → `qPL` | condition never true, rule never fires |
| v7364_m | `{tC_14.00, c0160}` → `qTU` | `eba_UE:x23` → `UE` (and `eba_qFI:qx2370`) | set can never match |
| v7368_m | `{tC_14.00, c0060}` → `qPO`, `{tC_14.00, c0061}` → `qST` | `eba_PL:x72` → `PL`, `eba_RT:x14` → `RT` | dead member in the `in` set; `!=` always true |
| v11649_m | `{tC_14.00, c0060}` → `qPO` | `eba_qSR:qx2057` → `qSR` | `!=` always true, precondition always fires |

Five of the nine compare against a pre-refit domain (`eba_AS`) where the cell now takes members of the refit domain (`qAS`), which is the recurring cause: domain renames leave legacy signatures behind in expressions that keep validating clean.

### Proposed Solution

```python
During semantic analysis, raise a warning when an item literal is compared against an enumerated component whose domain does not contain it.

Scope: `=`, `!=` and `in` (including each member of a set literal), wherever they occur — `where` clause, `if` condition, or the rule body — plus the value of a `sub` clause, which today accepts an out-of-domain item and silently yields an empty recordset.

Domain resolution, identical for a Fact Component and for an open key: the component's variable → its property → the property's category at the script's release, then the items of that category. There is no narrower notion of "possible values" available in the dictionary — the subcategory link on variable versions is unpopulated in 4.2.1, and the cell context describes which datapoint is addressed, not which values it may take — so the check is domain membership and nothing finer.

Rules to settle in the implementation:

- A selection can span several variables with different domains: warn only when the literal belongs to none of them, so a mixed selection never produces a false positive.
- In a set literal, warn once per impossible member, since a set can be partly dead (v7368_m).
- Resolve the domain at the release the script is validated against: some properties change domain between releases.
- Warning, not error: 4.2.1 ships operations that would trip it, and they must keep validating.
```

### Alternatives Considered

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Closes #332 

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/ast/domain_membership.py
+++ b/src/dpmcore/dpm_xl/ast/domain_membership.py
@@ -193,8 +193,11 @@ class DomainMembershipChecker(ASTTemplate):
         effect: str,
     ) -> None:
         reference, domains = component
-        item_domains = self._domains_of_items(literals)
-        for signature in literals:
+        # A set literal may repeat a member. The mismatch is one fact about
+        # the item, so it is reported once however often it is written.
+        signatures = list(dict.fromkeys(literals))
+        item_domains = self._domains_of_items(signatures)
+        for signature in signatures:
             found = item_domains.get(signature)
             # An item with no domain open at this release is reported as
             # not-found (1-1) before this pass runs; a set member the
--- a/src/dpmcore/dpm_xl/ast/domain_membership.py
+++ b/src/dpmcore/dpm_xl/ast/domain_membership.py
@@ -1,0 +1,331 @@
+"""Domain-membership check for comparisons against item literals.
+
+An enumerated component takes its values from exactly one *domain* (a
+``Category``). Comparing it with an item that belongs to a different domain is
+accepted by every other check -- the item exists (1-1) and ``Item`` is
+type-compatible with ``Item`` -- yet the comparison can never hold: ``=`` and
+``in`` are always false, ``!=`` is always true, and a ``sub`` clause on an
+out-of-domain item yields an empty recordset. The rule silently does something
+other than what it says.
+
+This pass reports that mismatch as a warning, not an error: EBA DPM 4.2.1
+ships operations that trip it (domain renames such as ``AS`` -> ``qAS`` leave
+legacy signatures behind in expressions that still validate clean), and those
+must keep validating.
+
+Domain resolution is the same for every component kind -- the component's
+property, then the category that property is typed on at the script's release.
+``Category.IsEnumerated`` is the gate: a property in a non-enumerated category
+(dates, identifiers, free text, "not applicable") describes no value set, so
+nothing is claimed about it. The dictionary offers nothing finer: the
+subcategory link on variable versions is unpopulated, so the check is domain
+membership and nothing more.
+
+The pass runs after semantic analysis has succeeded, so a genuine type error
+is reported on its own rather than alongside this warning.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING, Iterable
+
+from dpmcore.dpm_xl.ast.nodes import (
+    AST,
+    BinOp,
+    Dimension,
+    FilterOp,
+    GetOp,
+    ParExpr,
+    RenameOp,
+    Set,
+    SubAssignment,
+    SubOp,
+    TimeShiftOp,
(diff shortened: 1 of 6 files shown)


src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -314,6 +316,114 @@ def get_item_category_id_from_signature(
...
+            session: SQLAlchemy session.
+            items: Item signatures to resolve.
+            release_id: Release the membership is resolved at.
+
+        Returns:
+            ``{signature: {category_code, ...}}``, omitting signatures with
+            no category open at ``release_id``.
+        """
+        if not items:
+            return {}
```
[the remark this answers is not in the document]
victorprietosandamaso-cloud: I don't think this is worth changing the signature, passing a MagicMock() instead of None in the empty-items test does the same thing without weakening the contract for real callers. Not a blocker.

src/dpmcore/dpm_xl/ast/domain_membership.py
```diff
@@ -0,0 +1,331 @@
...
+            # An item with no domain open at this release is reported as
+            # not-found (1-1) before this pass runs; a set member the
+            # dictionary places in no enumerated category says nothing.
+            if not found or found & domains:
+                continue
+            add_semantic_warning(
+                f"Item [{signature}] belongs to domain {_join(found)}, "
+                f"but {reference} takes items from domain "
+                f"{_join(domains)}: {effect}."
+            )
```
ruizmaa: A set literal with a repeated out-of-domain member (e.g. `{...} in {[eba_PL:x72], [eba_PL:x72]}`) emits the same warning twice.

Is this the expected behavior or should it just warn once?

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -314,6 +316,114 @@ def get_item_category_id_from_signature(
...
+            release_id=release_id,
+            active_only_fallback=True,
+        )
+        domains: dict[int, set[str]] = {}
+        for row in query.distinct().all():
+            domains.setdefault(int(row.PropertyID), set()).add(
+                row.CategoryCode
+            )
+        return domains
+
```
ruizmaa, line 448: This function (`PropertyCategoryQuery.get_property_domains`) seems to do the same as `ItemCategoryQuery.get_item_domains`, same join Category, filters and grouping into a dict. The only differences are the model, the key column  and the release columns passed to filter_by_release.

Might be good to have a shared function for both models

src/dpmcore/dpm_xl/model_queries.py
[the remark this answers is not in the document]
andres-sole: I agree, resolving this one

src/dpmcore/dpm_xl/ast/domain_membership.py
andres-sole: it should warn once, fixed now.

src/dpmcore/dpm_xl/model_queries.py
andres-sole: Yes that is better. JUst added the shared helper on latest commit

ruizmaa (approved):

PR #335, opened 2026-08-26
fix: check if branches against the matched join structure rule

### LINKED ISSUE #333: check if branches against the matched join structure rule
### Initial Checks

### Summary

DPM-XL §8.1.5 states the constraint on key components, not on the scalar/recordset kind of the branches (*Matched join structure*): the key components of `then` and `else` must either be the same, or both be equal to (or a subset of) those of the condition. Equivalently, as §8.1.6 puts it, `join(condition, then)` and `join(condition, else)` must have the same structure.

Two consequences follow:

- with a **scalar** condition the constraint collapses to "both branches carry the same key components", so a scalar branch cannot be paired with a recordset one — this is the case `4-6-1-3` already covered, and the "if the condition is a scalar" qualifier in its message was correct;
- with a **recordset** condition the same pairing is **allowed**: the scalar branch is applied to every record of the condition. §8.1.6 lists it explicitly (`Recordset | Scalar | Recordset` and `Recordset | Recordset | Scalar`, result = the condition's key components), §8.1.7 example 6 is exactly that shape, and §8.1.6 presents `if A = x then y else A` as the general pattern for replacing a value.

So the absence of a kind check for a recordset condition is not the defect. The defects are the two below.

## Defect 1: a branch reaching outside the condition is accepted

When one branch carries a key component the condition does not have, and the other branch cannot supply values for it, §8.1.5 rejects the expression (§8.1.7 example 8: `if {tT1, r010-020} > 150 then 0 else {tT3, r020} endif`, where the joins have `{r}` and `{r, CNT}`). Semantic analysis accepts it: each branch is validated against the condition on its own, and the branch with the larger key component set simply becomes the result.

```python
from dpmcore import connect

# The condition is a single cell; the else branch spans two rows, so it adds
# a key component the then branch cannot fill.
EXPRESSION = (
    "if {tC_01.00, r0010, c0010} = 0 "
    "then 0.65 "
    "else {tC_17.01.a, (r0010, r0110), c0010} "
    "endif"
)

    result = db.services.semantic.validate(EXPRESSION, release_code="4.2.1")
    print(result.is_valid)  # True — expected False
```

This is also the shape in which the mismatch arises naturally with chained conditions, since a nested `if` over a row range returns a recordset while the outer branch is a literal:

```
if year({tC_04.00, r0900, c0010}[get refPeriod]) = 2028
then 0.65
else if year({tC_04.00, r0900, c0010}[get refPeriod]) = 2029
     then 0.70
     else 0.725
```

(Here every operand is scalar-like — a fully specified cell holds one value per global key combination, and §8.1.7 example 5 calls the comparison of such a cell a "Scalar Boolean" — so this particular expression is valid. It becomes invalid as soon as one branch of the chain spans a range of rows or columns.)

Expected: a semantic error reporting that the two branches do not produce the same key components once joined with the condition. Actual: validation succeeds and the expression is typed with the key components of whichever branch happens to carry more, which the other branch cannot populate.

## Defect 2: a set-valued branch surfaces as an internal error

The grammar accepts any expression in a branch, so `if <cond> then {1, 2} else {3} endif` parses, but the conditional operator only combines recordsets and scalars:

- with a recordset condition, both set branches are silently discarded and the condition's own structure becomes the result — including in the mixed case `then {1, 2} else 0.65`;
- with a scalar condition, analysis fails with an internal error instead of a semantic one, so the caller gets an internal error rather than a validation message.

Expected: a semantic error naming the offending branch.
In a branching `if-then-else`, the `then` and `else` branches can resolve to different kinds of result — a scalar in one branch and a recordset in the other — and semantic analysis accepts the expression without any error.

The mismatch is easy to hit with chained conditions, because a nested `if` whose condition is a cell selection returns a recordset, while a literal branch of the outer `if` returns a scalar:

```
if year({tC_04.00, r0900, c0010}[get refPeriod]) = 2028
then 0.65                                                     -- scalar (literal)
else if year({tC_04.00, r0900, c0010}[get refPeriod]) = 2029  -- recordset (nested if)
     then 0.70
     else 0.725
```

The outer `if` compares a scalar literal (`0.65`) against the recordset produced by the nested `if`, and no diagnostic is raised. The result reported for the whole expression is the recordset structure taken from the `else` branch, so the branch that "wins" depends on which side happens to be a recordset rather than on a validated agreement between the two branches.

There is already a semantic error for exactly this condition — *"For if then else operator, then and else have to be both recordset or both scalars"* (`4-6-1-3`) — but it is only reported when the condition of the `if` is a scalar. When the condition is a recordset, which is the usual case (any comparison over a cell selection, as above), the then/else kind mismatch is not checked at all.

## Proposed behaviour

Enforce the Matched join structure constraint of §8.1.5 for every `if`:

- compare the key components of `join(condition, then)` and `join(condition, else)`, and reject the expression when they differ (`4-6-1-3`, with its message stated in terms of key components and keeping the scalar-condition case as the special case it is);
- keep accepting a scalar branch paired with a recordset branch when the condition supplies the key components (§8.1.6, §8.1.7 examples 3 and 6);
- honour the Null literal exception: an omitted `else`, or a `null` branch, admits a wider other branch (§8.1.7 example 9);
- reject a set-valued branch with a semantic error.

### Reproducible Example
```python
from dpmcore import connect

# 1. The condition is a single cell; the else branch spans two rows, so it
#    carries a key component the then branch cannot fill (§8.1.7 example 8).
OUTSIDE_CONDITION = (
    "if {tC_01.00, r0010, c0010} = 0 "
    "then 0.65 "
    "else {tC_17.01.a, (r0010, r0110), c0010} "
    "endif"
)

# 2. A set in a branch, which the grammar accepts.
SET_BRANCHES = "if {tC_01.00, r0010, c0010} = 0 then {1, 2} else {3} endif"

# 3. For contrast: the shape the original report asked to reject, which the
#    specification allows (§8.1.7 example 6) and which must keep validating.
SCALAR_ELSE = "if {tC_17.01.a, (r0010, r0110), c0010} > 0 then 1 else 0 endif"

    for expression in (OUTSIDE_CONDITION, SET_BRANCHES, SCALAR_ELSE):
        result = db.services.semantic.validate(
            expression, release_code="4.2.1"
        )
        print(result.is_valid, result.error_message)
```

### Expected vs Actual

Expected, for the expression reaching outside its condition: invalid, with a conditional-operator error reporting that the two branches do not carry the same key components once joined with the condition.
Actual: `True`, no message. The expression is typed with the `r` key component of the `else` branch, which the scalar `then` branch cannot populate.

Expected, for the set-valued branches: invalid, with a semantic error naming the offending branch.
Actual: an internal error (`AttributeError`) surfaces to the caller under a scalar condition; under a recordset condition the branches are dropped and the condition's structure is returned as the result.

Expected, for the scalar `else` under a recordset condition: valid, typed with the key components of the condition.
Actual: valid. No change is due here — this is recorded because the original description of this issue asked for it to be rejected, which §8.1.6 and §8.1.7 example 6 do not support.

### dpmcore version

0.1.2rc10

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Fixes #333

DPM-XL §8.1.5 constrains the `then`/`else` branches of an `if` by their key components, not by their scalar/recordset kind (*Matched join structure*): the key components of the two branches must either be the same, or both be equal to (or a subset of) those of the condition. §8.1.6 states the same thing as a requirement on the joins — `join(condition, then)` and `join(condition, else)` must have the same structure.

Semantic analysis checked something else. With a scalar condition it required the two branches to be both recordsets or both scalars, which is the correct specialisation of the rule for a condition with no key components of its own. With a recordset condition it validated each branch against the condition separately and then took whichever branch had the larger key component set, so an expression whose branches disagreed on their key components was accepted and typed with a key component one branch can never populate — the case §8.1.7 example 8 rejects.

The check is now the rule itself: the key components of `join(condition, then)` and `join(condition, else)` are compared, and `4-6-1-3` is raised when they differ. Consequences:

- a branch reaching outside the condition while the other cannot follow it there is rejected — the defect behind this issue;
- a scalar branch paired with a recordset branch stays valid under a recordset condition, where the scalar is applied to every record of the condition (§8.1.6 rows `Recordset | Scalar | Recordset` and `Recordset | Recordset | Scalar`, §8.1.7 examples 3 and 6, and the `if A = x then y else A` replacement pattern of §8.1.6);
- both branches inside a wider condition are valid even when they disagree with each other, since each joins with the condition to the same structure — case (b) of §8.1.5, which the previous subset comparison between the two branches could not express;
- the Null literal exception is honoured: an omitted `else`, or a `null` branch, admits an other branch wider than the condition (§8.1.7 example 9). A `null` branch used to be rejected as a kind mismatch;
- a set-valued branch (`then {1, 2}`), which the grammar accepts, raises the new semantic error `4-6-1-2` instead of being dropped on the floor or surfacing as an internal `AttributeError`.

A fully specified cell — a recordset whose key components are all global — counts as scalar-like, as it holds one value per global key combination and contributes no key component a scalar lacks. The specification reads it the same way, calling the comparison of such a cell a "Scalar Boolean" (§8.1.7 example 5). This is the decision the issue left open; it is what the scalar-condition path already did, and it keeps the chained `else if` from the report valid.

## Checklist

## Impact / Risk


victorprietosandamaso-cloud (dismissed):

victorprietosandamaso-cloud (dismissed):

src/dpmcore/dpm_xl/operators/conditional.py
ruizmaa: Same bug this PR just fixed for `IfOperator`. nvl's spec (§8.2.2) says `op1`/`op2` must be `rset | scal <*>`. A ScalarSet operand (e.g. `{1,2}`) is not one of those. It still falls through `check_structures` to `None`, `None` and later raises an exception in validate instead of a `SemanticError`

```python
import dpmcore.dpm_xl.semantic_analyzer
from dpmcore.dpm_xl.operators.conditional import Nvl
from dpmcore.dpm_xl.symbols import Scalar, ScalarSet
from dpmcore.dpm_xl.types.scalar import Number

left = ScalarSet(type_=Number(), name="my_set", origin="my_set")
right = Scalar(type_=Number(), name="x", origin="x")
Nvl.validate(left, right)
# Exception: Nvl produced no result structure; unhandled operand combination
```

Should we fix this in the same PR or open a new issue to manage it?

andres-sole: Yes i didn't notice. Fixed it on this PR and added test for nvl op.

ruizmaa (approved):

PR #341, opened 2026-08-27
fix(ast-generator): walk the gate AST so parameters reach the engine

### LINKED ISSUE #338: Parameters in a precondition expression are accepted but never reach the generated script
### Root cause

  `_build_preconditions_block` extracts precondition items via regex over the gate string (`_VAR_REF_PATTERN` matching `{v_*}`), and never touches the `{p_*}` positions. Everything else — logical operators, negations,grouping — is discarded too, and a parameter-only gate emits nothing atall, which is why the gate silently vanishes.

  ### Fix

  Replace the regex path with a walk over the gate's AST. The gate string is parsed with `self._syntax.parse(prec["expression"])` (already available in the pipeline), and the walker emits:
  - `VarID` with a `v` prefix → `PreconditionItem` with the resolved
    `variable_id` / `variable_vid`;
  - `ParameterRef` → the same node shape expression parameters emit, with
    `code` and `declared_type` preserved; the parameter is added to the
    script-level `parameters` registry in the same pass, so scope-wide type
    agreement runs against a single source of truth;
  - `BinOp` (`and`, `or`, ...) / `UnaryOp` (`not`) / `ParExpr` → preserved
    as their AST nodes, so `{v_A} or {v_B}` no longer collapses to two
    disconnected items and `{v_A} and not {v_B}` no longer loses the
    negation.
   
  The public shape of `script()` output is unchanged — same keys, same `preconditions[<key>]` layout for gates that today only carry filing ndicators. What changes is that the tree under `ast` now reflects what the gate actually says, and gates that mix items with parameters — or consist of a parameter alone — reach the engine intact.

  ### Notes on edge cases
  - Gate parameter-only (`{p_flag, boolean}`): emits a precondition entry with a `ParameterRef` AST and registers `flag: Boolean` at cript-level. No more silently ungated validations.
  - Gate mixing item + parameter (`{v_F_01.02} and {p_flag, boolean}`): emits a `BinOp(and, PreconditionItem, ParameterRef)`.
  - Same parameter code declared with two different `declared_type`s across a script (whichever half of the pair — precondition or expression) is still rejected by the existing scope-wide agreement check, since both now populate the same registry.
  - Codes that don't resolve in the DB continue to be skipped silently, astoday.
   
  ### Follow-ups (out of scope)

  - Merge policy for gates that produce the same canonical AST across operations is unchanged (`_merge_precondition_entry`) but the key derivation may need revisiting if a gate has 0 items.

victorprietosandamaso-cloud:
## Summary

  Fixes #338

  ## Problem

  `_build_preconditions_block` extracted precondition items with a regex over the gate source text that only matched `{v_*}` positions. Anything the regex could not name — a `{p_*}` parameter reference, a logical operator connecting one to a filing indicator, a negation, a `ParExpr` — was dropped on the floor.

  A gate like `{v_F_01.02} and {p_flag, boolean}` reached the engine as a bare `PreconditionItem` for `F_01.02`, and a parameter-only gate `{p_flag, boolean}` reached the engine as nothing at all, silently ungating the operation.

  ## Fix

  Replace the regex path with a walker over the parsed gate AST. The gate is parsed once with `self._syntax.parse(prec["expression"])` (already available in
  the pipeline), and the walker emits:
  - `VarRef` (the parser's node for `{v_*}`) and `VarID` positions with the `v` prefix collapse to the `PreconditionItem` dict shape the engine binds against.
  - `ParameterRef` is preserved with `code` and `param_type` intact, and the parameter is added to the script-level `parameters` registry through the same
  `merge_parameters` used by expressions — scope-wide type agreement (`3-8`) now runs against a single source of truth even when the parameter only appears in a gate.
  - `BinOp` (`and` / `or` / ...), `UnaryOp` (`not`) and `ParExpr` are preserved as their AST nodes, so `{v_A} or {v_B}` no longer collapses to a flat item list under an implicit `and` and `{v_A} and not {v_B}` no longer loses the negation.
  - Every other node kind delegates to the standard `serialize_ast` path, so the tree the engine sees matches the shape used in expressions.

  ## Backward compatibility

  The public shape of `script()` output is unchanged for gates that used to work:

  - `preconditions[<key>]` still groups entries under `p_<sorted_vids>` when the gate names at least one filing indicator, so gates that today merge two
  operations under one key keep doing so.
  - `precondition_variables` still tracks the resolved variable vids.
  - `_build_preconditions_block` runs before `parameters_block` is finalised, so gate parameters land in the same registry as expression parameters without changing the block's public keys.

  Gates that carry no filing indicator (a parameter-only gate) derive their key from a `zlib.crc32` of the emitted AST — a stable slot instead of overwriting
  the previous one.
## Tests

  New `TestGateParameterPropagation` covers the three shapes from the issue:

  - parameter-only gate emits a `ParameterRef` entry and populates the script-level `parameters` registry;
  - mixed gate `{v_F_01.02} and {p_flag, boolean}` reaches the engine as a `BinOp` whose operands are `PreconditionItem` + `ParameterRef`, both sides intact;
  - `{v_A} or {v_B}` preserves the `or` operator instead of collapsing to a flat item list.

  The pre-existing `TestBuildPreconditionsBlock` tests are updated for the new AST-based shape but keep asserting the same `p_<vid>` naming, the same merge-on-collision behaviour, and the same dict-vs-tuple input handling.

  Ruff + mypy clean; the three pre-existing failures in `test_ast_generator.py` (missing `release_sort_order` stub, missing `rich`) are unchanged against master.

  ## Checklist
  - [x] Tests pass (`pytest`) — 1833 unit tests passing, no regressions against master

  ## Impact / Risk

  - Breaking changes? No public API change. Gate outputs for pre-existing shapes are byte-compatible; gates that carry a parameter or non-`and` operator now
  reach the engine differently — that is the fix.
  - Database schema or migration concerns? None.
  - Notes for release/changelog? Yes — call out that gates with `{p_*}` references, `or`, `not`, or grouping now reach the engine intact.

### DIFF
--- a/src/dpmcore/services/ast_generator.py
+++ b/src/dpmcore/services/ast_generator.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import logging
 import re
 import zlib
@@ -43,7 +44,6 @@ if TYPE_CHECKING:
 logger = logging.getLogger(__name__)
 
 
-_VAR_REF_PATTERN = re.compile(r"\{v_?([^}]+)\}")
 _TABLE_CODE_NORMALIZER = re.compile(r"^([A-Z]+)_(\d+)_(\d+)$")
 _DEFAULT_FROM_DATE = "2001-01-01"
 _DEFAULT_NAMESPACE = "default_module"
@@ -314,21 +314,27 @@ class ASTGeneratorService:
             for tbl in tables_block.values():
                 variables_block.update(tbl.get("variables", {}))
 
+            preconditions_block, precondition_variables_block = (
+                self._build_preconditions_block(
+                    preconditions or [],
+                    release_id=release_id,
+                    referenced_parameters=referenced_parameters,
+                )
+            )
+
             # Runtime-binding contract: the declared type of every parameter
             # this script's operations reference, keyed by code. This is the
             # scope-wide invariant. ``is_set`` is recoverable from the ``set-``
             # prefix and ``default`` is a per-reference fallback the engine
-            # binds per scope, so neither belongs in this registry.
+            # binds per scope, so neither belongs in this registry. Populated
+            # after ``_build_preconditions_block`` runs so parameters that
+            # appear only in a gate — legal per the grammar, and reaching the
+            # engine as a runtime input — are declared alongside the ones the
+            # expressions reference.
             parameters_block: Dict[str, str] = {
                 prm_code: prm.declared_type
                 for prm_code, prm in sorted(referenced_parameters.items())
             }
-
-            preconditions_block, precondition_variables_block = (
-                self._build_preconditions_block(
-                    preconditions or [], release_id=release_id
-                )
-            )
 
             dependency_info = self._build_dependency_info(
                 scope_pairs=scope_pairs,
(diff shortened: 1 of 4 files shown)


src/dpmcore/services/ast_generator.py
```diff
@@ -1345,101 +1366,458 @@ def _merge_precondition_entry(
...
+            serialized: Dict[str, Any] = serialize_ast(node)
+            return serialized
+        if isinstance(node, ast_nodes.BinOp):
+            left = cls._transform_precondition_ast(
+                node.left, resolved, precondition_variables, serialize_ast
+            )
+            right = cls._transform_precondition_ast(
+                node.right, resolved, precondition_variables, serialize_ast
+            )
+            if left is None:
```
guillermo-garcia-1: and/or/xor all handle an unresolved operand the same way: drop it, keep the other side. That's fine for and (dropping a condition just relaxes it, matches existing pydpm behavior). But {v_A} or {v_B} with A unresolved silently becomes just {v_B} — a stricter condition, not a looser one. For xor it makes even less sense. No test covers this case.

andres-sole: Fixed all three and added tests

guillermo-garcia-1 (approved):

PR #351, opened 2026-09-03
fix: serialize year/month/day on DateConstructorOp

### LINKED ISSUE #350: `DateConstructorOp` serializer silently drops `year`/`month`/`day` (and uses an undecided `class_name`)
### Initial Checks

### Summary

`DateConstructorOp` (the `date(year, month, day)` constructor) has no dedicated visitor in the JSON serializer, so it falls through to `generic_visit`. This causes two problems at once: the serialized `class_name` is the raw Python class name rather than a decided wire name, and, much more seriously, `generic_visit` only copies a fixed, generic set of attributes that does not include `year`, `month` or `day` — those three fields are silently dropped from every serialized `date(...)` node.

### Reproducible Example

```python
from dpmcore.dpm_xl.utils.serialization import serialize_ast

ast = SyntaxService().parse("date(2020, 1, 1)")
node = ast.children[0]
print(type(node).__name__)          # DateConstructorOp
print(serialize_ast(node))
```

### Expected vs Actual

Expected: the serialized node carries `year`, `month` and `day`, each serialized as a Node (e.g. `Constant`), matching how every other multi-argument operator (`SubstrOp`, `AnnualiseOp`, `TimeShiftOp`) is serialized by its own dedicated `visit_*` method.
Actual: `{'class_name': 'DateConstructorOp', 'op': 'date'}`

### dpmcore version

0.1.2rc10

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

ruizmaa:
Closes #350 

## Summary

`date(year, month, day)` silently lost `year`/`month`/`day` when serialized through the real path (`serialize_ast` / `ASTToJSONVisitor`), because `DateConstructorOp` had no dedicated `visit_DateConstructorOp` and fell back to the generic `generic_visit`, which doesn't cover those fields. The only existing test missed it because it called `node.toJSON()`.

This PR adds the missing `visit_DateConstructorOp` (serializing `year`/`month`/`day` the same way `visit_AnnualiseOp`/`visit_SubstrOp` do) plus a test against the real path that fails without the fix.

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/utils/serialization.py
+++ b/src/dpmcore/dpm_xl/utils/serialization.py
@@ -713,6 +713,21 @@ class ASTToJSONVisitor(NodeVisitor):
             "reference_period": node.component,
         }
 
+    def visit_DateConstructorOp(self, node: Any) -> NodeDict:
+        """Visit DateConstructorOp nodes.
+
+        ``year``/``month``/``day`` are full Nodes (§10.3.5 admits a
+        Recordset in any of the three positions), so each is serialized
+        recursively rather than copied as a raw value.
+        """
+        return {
+            "class_name": "DateConstructorOp",
+            "op": node.op,
+            "year": self.visit(node.year),
+            "month": self.visit(node.month),
+            "day": self.visit(node.day),
+        }
+
     def visit_AnnualiseOp(self, node: Any) -> NodeDict:
         """Visit AnnualiseOp nodes."""
         return {
--- a/tests/unit/ast/test_date_constructor_expression.py
+++ b/tests/unit/ast/test_date_constructor_expression.py
@@ -13,6 +13,7 @@ from dpmcore.dpm_xl.symbols import (
     Structure,
 )
 from dpmcore.dpm_xl.types.scalar import Date, Integer, Mixed, Number, String
+from dpmcore.dpm_xl.utils.serialization import ASTToJSONVisitor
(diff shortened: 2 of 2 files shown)


andres-sole (commented):
This doesn't match what the engine deserializes and a missing field fails the whole script file

Current:

```json
{ "class_name": "DateConstructorOp", "op": "date", "year": {...}, "month": {...}, "day": {...} }
```

Accepted:

```json
{ "class_name": "DateConstructorOp", "year_node": {...}, "month_node": {...}, "day_node": {...} }
```


Are you coordinating with the engine team? Whats the intended shape? @ruizmaa

ruizmaa:
Hi @andres-sole , I haven't been in touch with the engine team. I was just checking the script format specifications Antonio is writing in `dpm-xl-docs` ([MR !77](https://gitlab.com/aollerosr/dpm-xl-docs/-/merge_requests/77)). While reviewing this I checked what §4.16 requires for `DateConstructorOp` and I realized dpmcore's serialization didn't match it, it was silently dropping year/month/day entirely (#350).

I also checked `cve`, and it already expects exactly this shape (year/month/day)

So the point of this PR is to align dpmcore with this new spec, which is expected to be soon the source of truth of the specification. I guess that the engine should also stick to this spec, so every project uses the same serialization and follows the same rules.

andres-sole (approved):

PR #352, opened 2026-09-03
Feat/layout export enumeration values

antonio-olleros:
## Summary

Several fixes to the Excel Export to be deployed in PROD to the DPM Studio

## Checklist

## Impact / Risk

- Breaking changes? (public API / CLI / REST endpoints / Django models) no
- Database schema or migration concerns? no
- Notes for release/changelog? no

### DIFF
--- a/src/dpmcore/services/layout_exporter/excel_writer.py
+++ b/src/dpmcore/services/layout_exporter/excel_writer.py
@@ -18,6 +18,8 @@ from openpyxl.worksheet.hyperlink import Hyperlink
 from dpmcore.services.layout_exporter.models import (
     CellData,
     DimensionMember,
+    Enumeration,
+    EnumValue,
     ExportConfig,
     LayoutHeader,
     TableLayout,
@@ -53,6 +55,16 @@ _MAX_SHEET_TITLE = 31
 # Cap the identity list in a tooltip; a few datapoints are reported in
 # dozens of cells and the comment box would be unreadable.
 _MAX_IDENTITIES_IN_TOOLTIP = 15
+# Excel rejects a comment longer than 32,767 characters and shows a
+# repair dialog for the whole workbook, so the list of possible values
+# of a very large hierarchy (NACE runs to 841 codes) is cut short.
+_MAX_TOOLTIP_CHARS = 30000
+# Comment box geometry: it grows with the text so a long value list is
+# readable without resizing, up to a height that still fits a screen.
+_COMMENT_WIDTH = 400
+_COMMENT_MIN_HEIGHT = 200
+_COMMENT_MAX_HEIGHT = 600
+_COMMENT_LINE_HEIGHT = 14
 
 _TITLE_FONT = Font(bold=True, size=11)
 _HEADER_FONT = Font(bold=True, size=10)
@@ -356,11 +368,19 @@ class ExcelLayoutWriter:
                     ann_val_cell.font = Font(size=9)
                     ann_val_cell.alignment = Alignment(horizontal="left")
 
-                if cfg.add_header_comments and sh.categorisations:
-                    tooltip = _format_categorisations(sh.categorisations)
-                    sc.comment = Comment(
-                        tooltip, "dpmcore", width=400, height=150
-                    )
+                if cfg.add_header_comments:
+                    # An open-sheet table keys its sheets off the Z
+                    # header's variable, so its possible values belong
+                    # here — there is no key cell to hang them on.
+                    parts = []
+                    if sh.categorisations:
+                        parts.append(
+                            _format_categorisations(sh.categorisations)
+                        )
+                    if sh.is_key and sh.key_variable_vid:
+                        parts.append(_key_cell_tooltip(sh))
+                    if parts:
+                        sc.comment = _comment("\n\n".join(parts))
 
         # --- "Columns" label ---
         col_cell = ws.cell(
@@ -616,11 +636,8 @@ class ExcelLayoutWriter:
 
                     # Cell tooltip
                     if cfg.add_cell_comments:
-                        cell.comment = Comment(
+                        cell.comment = _comment(
                             _cell_tooltip(cell_data, others),
-                            "dpmcore",
-                            width=400,
-                            height=200,
                         )
 
         # --- Annotations ---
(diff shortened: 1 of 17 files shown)


src/dpmcore/services/layout_exporter/processing.py
```diff
@@ -202,6 +206,136 @@ def build_cells(
     return cells
 
 
+def derive_missing_cell_data(
```
andres-sole, line 209: this contradicts itself on axis precedence. The order is stated to be table, sheet, row and column but its column, row and sheet

src/dpmcore/services/layout_exporter/excel_writer.py
```diff
@@ -162,11 +181,19 @@ def _write_index(self, entries: list[_IndexEntry]) -> None:
         # A "Sheet" column is only meaningful when at least one table
         # was split into one worksheet per Z-axis sheet.
         has_sheets = any(e.sheet_label for e in entries)
+        # So is a "Pending" one: it counts the cells still waiting for
+        # a variable, which is zero for a finished dictionary.
+        pending = {
+            id(e.layout): _count_pending_cells(e.layout) for e in entries
+        }
```
ruizmaa: `id(e.layout)` identifies the table layout, but not the sheet. Since all sheets of a Z-axis-split table share the same `TableLayout`, all Index entries for that table end up with the same pending-cell count

src/dpmcore/services/layout_exporter/excel_writer.py
```diff
@@ -1057,12 +1069,75 @@ def _build_identity_index(layouts: list[TableLayout]) -> dict[int, list[str]]:
...
+        for (_row_id, _col_id, cell_sheet_id), cd in layout.cells.items()
+        if cell_sheet_id == sheet_id
+        and not cd.variable_vid
+        and not cd.is_excluded
+        and not cd.is_void
+    )
+    # The key cells of an open table are rendered from their header
+    # and have no ``cells`` entry of their own, so they have to be
+    # counted here too — the worksheet already flags them as pending.
+    if not layout.rows:
```
guillermo-garcia-1, line 1104: This only accounts for open-row tables' pending key cells. An open-sheet table's Z key header can also be pending (no key_variable_vid), and its tooltip already flags it as such, but it's never added to this count — so a table whose only incompleteness is its Z key can show pending == 0, and if that holds workbook-wide the "Cells without a variable" index column disappears entirely. Should this also walk layout.sheets?

src/dpmcore/services/layout_exporter/queries.py
```diff
@@ -75,6 +89,53 @@ def load_module_table_versions(
     return q.all()
 
 
+def load_module_version(
+    session: Session,
+    module_code: str,
+    release_code: Optional[str] = None,
+) -> Optional[Any]:
+    """Load the ModuleVersion whose release window the export uses.
```
guillermo-garcia-1: load_module_table_versions unions every matching ModuleVersion row; this picks one via "latest-starting wins." When a draft (started at the undated/perpetual release) coexists with an already-adopted version, that tie-break picks the draft — the opposite of what _resolve_table_version_scope already does for table versions in the same situation. If this is reachable for module versions too, it needs the same "prefer adopted" fix, not just a docstring update.

src/dpmcore/services/layout_exporter/excel_writer.py
```diff
@@ -1074,9 +1149,96 @@ def _cell_tooltip(cell_data: CellData, identities: list[str]) -> str:
...
+    # The marker counts against the cap too, so reserve its longest
+    # possible form up front. Without this the result overshoots by the
+    # marker's length, which would also make a second pass over an
+    # already-fitted tooltip truncate it again.
+    budget = _MAX_TOOLTIP_CHARS - len(f"... and {len(lines)} more lines\n")
+    kept: list[str] = []
+    used = 0
+    for index, line in enumerate(lines):
+        used += len(line) + 1
+        if used > budget:
```
guillermo-garcia-1, line 1230: If a single line alone exceeds _MAX_TOOLTIP_CHARS , it's dropped entirely on the first iteration and the tooltip shows only the "... and N more lines" marker — no real content at all. Is a single line ever expected to get anywhere close to that length in practice?

src/dpmcore/services/layout_exporter/processing.py
```diff
@@ -202,6 +206,141 @@ def build_cells(
...
+    keeps its last, most specific member.
+    """
+    merged: dict[int, DimensionMember] = {}
+    for group in groups:
+        for dm in group:
+            merged[dm.property_id] = dm
+    return list(merged.values())
+
+
+def resolve_enumeration_sources(
```
guillermo-garcia-1, line 288: This assumes exactly one of column/row/sheet carries a subcategory_vid for a given enumerated cell. If that's ever violated it silently prefers column > row > sheet with no validation or log. Intentional?

guillermo-garcia-1:
This PR adds a user-facing CLI flag (--derive-missing-variables/--no-derive-missing-variables) and two new visible export behaviors (possible values in enumerated-cell comments, pending-cell shading/index column), but docs/cli.rst wasn't updated to cover any of them.

andres-sole:
@guillermo-garcia-1 Added the flag info to the docs

src/dpmcore/services/layout_exporter/excel_writer.py
andres-sole: fixed in latest commits

src/dpmcore/services/layout_exporter/queries.py
andres-sole: not reachable in published data (no module code matches twice at any release of 4.2.1 or 4.3) but fixed de error in the union regardless

src/dpmcore/services/layout_exporter/excel_writer.py
andres-sole: Not ocurring in published data. Fixed nevertheless by cutting the line instead of dropping

src/dpmcore/services/layout_exporter/processing.py
andres-sole: intentional. Added some docstrings to clarify

guillermo-garcia-1 (dismissed):

guillermo-garcia-1 (approved):

PR #358, opened 2026-09-09
fix: apply the ghost-module-version fallback to table-version resolution

### LINKED ISSUE #356: Table-version resolution ignores the ghost-module-version fallback
### Initial Checks

### Summary

At releases 4.0 and 4.1 a cell selection resolves its datapoints through the table version of a *ghost* module version, so the variables, properties and domains it reports are the ones the ghost introduced rather than the ones the effective module version actually uses. dpmcore already skips ghost module versions when resolving modules, so the two resolutions contradict each other for the same release.

Surfaced as a false-positive domain-membership warning: `{tF_40.01, c0095} = [eba_CT:x12]` is reported at release 4.0 as

```
Item [eba_CT:x12] belongs to domain CT, but { tF_40.01, c0095 } takes items from domain qSR: the comparison is never true.
```

which is wrong at 4.0 — the column still takes `CT` items there.

### Reproducible Example

```python
db.services.semantic.validate(
    "{tF_40.01, c0095} = [eba_CT:x12]", release_code="4.0"
)   # warns; should be clean


The two resolutions side by side, for table `F_40.01`:

| release | module resolution (ghost fallback) | datapoint resolution | domain of `c0095` |
| --- | --- | --- | --- |
| 4.0 | FINREP9 3.1.0 | table version of FINREP9 **3.2.0** | `qSR` (should be `CT`) |
| 4.1 | FINREP9 3.1.0 | table version of FINREP9 **3.2.0** | `qSR` (should be `CT`) |
| 4.2+ | FINREP9 3.3.0 | table version of FINREP9 3.3.0 | `qSR` (correct) |

FINREP9 3.2.0 is a ghost: its reference-date window is collapsed (`2024-12-31` on both bounds) and it sits inside 3.1.0's window, which runs to `2026-03-30` and joins 3.3.0's start at `2026-03-31`. It contributes no reporting period, but its release window (4.0 → 4.2) is what the table-version filter lands on.

The header behind `c0095` ("Sector of investee") is retyped from a `CT`-typed property to a `qSR`-typed one at release 4.0, and that retyping reaches the dictionary only through the ghost's table version. 3.1.0's table version keeps the `CT`-typed variable; 3.3.0's carries the `qSR`-typed one. So the retyping is real from 4.2 onward and the warning is correct there — it is a false positive only across the ghost's window.
```

### Expected vs Actual

Expected: when the only module version whose release window covers the target release is a ghost, datapoint and table-version resolution fall back to the latest prior non-ghost version of the same module, exactly as module resolution already does. At 4.0/4.1 that is FINREP9 3.1.0, and `{tF_40.01, c0095}` resolves to the `CT`-typed variable.

Actual: table-version resolution filters on the raw release window with no ghost awareness, so it selects the ghost's table version and every downstream answer (variable id, property, domain) comes from the ghost.

### dpmcore version

0.1.2rc12

### Python version

_No response_

### Database backend

_No response_

### OS

_No response_

### Additional Context

_No response_

andres-sole:
## Summary

Closes #356 

## Checklist

## Impact / Risk

### DIFF
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -1286,6 +1286,17 @@ def _ghost_module_ids_for_table_vids(
         ``table_vids`` (the table version is then genuinely live at the
         release and needs no substitution). Also ``None`` when no module
         version hosts them at all.
+
+        The decision is all-or-nothing over ``table_vids``: a live host
+        for *any* of them suppresses substitution for the whole set.
+        That only matters when one table code has several versions open
+        at once, which happens at the perpetual release alone (see
+        :meth:`ViewDatapointsQuery._resolve_table_version_scope`), and
+        deciding per version would also have to merge the live module
+        versions into ``_TableVersionScope.fallback_module_vids`` --
+        :meth:`ViewDatapointsQuery._filter_module_versions` replaces the
+        release filter with that list, so a live version left out of it
+        would lose its rows.
     """
     rows = (
         filter_by_release(
@@ -1331,6 +1342,14 @@ def _apply_table_ghost_fallback(
     resolution report the same module version. Without it the cells --
     hence their variables, properties and domains -- come from a version
     that has no reporting period of its own.
+
+    Kept separate from that helper rather than calling it: the caller
+    needs a scope (the effective ``table_vids`` *and* the module
+    versions the ``ModuleVersion`` join must be narrowed to), not the
+    operand rows the helper returns; a ghost with no fallback is kept
+    here instead of dropped, since dropping it would leave ``table``
+    unresolvable rather than merely unscoped; and the decision is made
+    once per table code rather than per module.
 
     Substitution only: when there is nothing to fall back to (no prior
     non-ghost version, or one that does not contain ``table``) the ghost's
--- a/src/dpmcore/dpm_xl/model_queries.py
+++ b/src/dpmcore/dpm_xl/model_queries.py
@@ -15,6 +15,7 @@ from typing import (
     Callable,
     Collection,
     Hashable,
+    NamedTuple,
     Sequence,
 )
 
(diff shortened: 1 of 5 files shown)


src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -1238,6 +1239,149 @@ def _resolve_with_ghost_fallback(
...
+
+def _apply_table_ghost_fallback(
+    session: "Session",
+    table: str,
+    table_vids: list[int],
+    release_id: int,
+) -> _TableVersionScope:
+    """Substitute a ghost-only table version with its fallback's (#356).
+
+    The table-version mirror of :func:`_resolve_with_ghost_fallback`: when
```
guillermo-garcia-1, line 1338: The docstring says this mirrors _resolve_with_ghost_fallback but not why it's separate rather than calling it directly — is it because table_vids here is already narrowed by upstream "adopted version" logic that get_from_table_codes doesn't account for?

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -1238,6 +1239,149 @@ def _resolve_with_ghost_fallback(
...
+            start_col=ModuleVersion.start_release_id,
+            end_col=ModuleVersion.end_release_id,
+            release_id=release_id,
+        )
+        .distinct()
+        .all()
+    )
+    ghost_module_ids: set[int] = set()
+    for module_id, from_date, to_date in rows:
+        if not _is_collapsed_window(from_date, to_date):
```
guillermo-garcia-1, line 1323: If table_vids has multiple candidates and one is ghost-only while another is live, this returns None for the whole set on the first live hit — so the ghost-only one never gets substituted. Reachable in practice, or are the two cases mutually exclusive?

guillermo-garcia-1 (dismissed):

src/dpmcore/dpm_xl/model_queries.py
andres-sole: Not reachable on published dictionaries. This unreproducible case have a large blast radious to fix. SO it would be worth noting in a comment for later development. Added an extension o docstring comment.

src/dpmcore/dpm_xl/model_queries.py
andres-sole: Added an extension o docstring comment. Not the adopted-version narrowing that stays upstream in `_resolve_table_version_scope` and is ortogonal to the fallback

guillermo-garcia-1 (approved):
I noticed Copilot's comment on hierarchy.py about the pandas import — might be worth a quick look at some point, but since it was marked as resolved, I'll trust that call and approve.

PR #376, opened 2026-09-17
perf: read a table's cells once per release instead of once per cell selection

### LINKED ISSUE #361: SemanticService.validate() spends ~31s of DB CPU per call in get_axis_orders / get_table_data
### Initial Checks

### Summary

Validating real DPM-XL expressions is dominated by a small family of queries in
`dpm_xl/model_queries.py`. The Python side is idle while they run: during a call the
process sits at ~0.8% CPU while SQL Server saturates 10+ cores (~1045%).

From the SQL Server plan cache after ~1 h of validation (statements touching `TableVersionCell`):

| executions | mean CPU | mean logical reads |
|---:|---:|---:|
| 94 | 31.5 s | 1,850,469 |
| 72 | 16.5 s | 1,297,345 |
| 61 | 6.7 s | 3,412,668 |

Millions of pages scanned to return a handful of row/column/sheet codes. Wait type `CXPACKET`.
Over a workload of ~7,400 expressions this adds up to hours of database CPU.

The statement is a `SELECT DISTINCT` over a 9-table join with no selective predicate:

    SELECT DISTINCT HeaderVersion_1.Code AS row_code, HeaderVersion_2.Code AS column_code, ...
    FROM TableVersion
    JOIN ModuleVersionComposition ... JOIN ModuleVersion ...
    JOIN TableVersionCell ... LEFT JOIN VariableVersion ... LEFT JOIN Property ...
    LEFT JOIN DataType ... JOIN Cell ... LEFT JOIN TableVersionHeader x3

Call path:

    SemanticService.validate()           services/semantic.py:234
      -> semantic_analyzer.py:180,188 / ast/operands.py:515
           -> ViewDatapointsQuery.get_table_data()   model_queries.py:1896
                -> get_axis_orders()                 model_queries.py:1798  (.distinct() L1842)
      -> ast/module_dependencies.py:162
           -> get_filtered_datapoints()              model_queries.py:2007  (.distinct() L2045)

A second factor: the SQL is built with a variable number of bound parameters (`@P1…@P48`),
so there is no plan reuse — 311 distinct cached plans for one logical query.

### Reproducible Example

```python
import time
from dpmcore import connect

# Real EBA validation expressions. Synthetic short ones do NOT reproduce this.
EXPRESSIONS = [
    "{v_C_06.02} or {v_C_105.01} or {v_C_105.03} or {v_C_27.00} or {v_C_34.06} or "
    "{v_C_67.00} or {v_C_71.00} or {v_C_77.00} or {v_F_40.01} or {v_I_11.03} or "
    "{v_T_07.00} or {v_T_08.00} or {v_T_12.00} or {v_Z_05.01} or {v_Z_05.02}",

    "{v_C_28.00} or {v_I_07.00} or {v_I_08.01} or {v_I_08.02} or {v_I_08.03} or "
    "{v_I_08.04} or {v_I_08.05} or {v_I_08.06} or {v_T_12.00} or {v_Z_05.01}",

    "with {tC_01.00, c0010, default: 0, interval: true}: "
    "{r0030} = {r0040} + {r0060} + {r0070} + {r0092}",
]

RELEASE_ID = 1010000115  # the release the expressions belong to

with connect("mssql+pyodbc://...") as db:
    for expr in EXPRESSIONS:
        t = time.perf_counter()
        result = db.services.semantic.validate(expr, release_id=RELEASE_ID)
        print(f"{time.perf_counter() - t:7.2f}s  valid={result.is_valid}  {expr[:60]}...")

While this runs, the cost is visible on the server side:


SELECT qs.execution_count,
       qs.total_worker_time / qs.execution_count / 1000   AS mean_cpu_ms,
       qs.total_logical_reads / qs.execution_count        AS mean_logical_reads,
       SUBSTRING(st.text, CHARINDEX('SELECT', st.text), 200) AS sql_text
FROM sys.dm_exec_query_stats qs
CROSS APPLY sys.dm_exec_sql_text(qs.sql_handle) st
WHERE st.text LIKE '%TableVersionCell%'
ORDER BY qs.total_worker_time DESC;
```

### Expected vs Actual

Expected: validating an expression resolves its table axes with queries proportional to the
tables referenced — a few thousand logical reads, single-digit milliseconds of database CPU.

Actual: each call to get_axis_orders / get_filtered_datapoints costs 6-31 s of database CPU
and scans 1.3-3.4 million pages, to return a handful of row/column/sheet codes. The Python
process is idle throughout (~0.8% CPU) while SQL Server saturates 10+ cores. Because the SQL
is built with a variable parameter count, 311 separate plans are cached for the same logical
query, so plans are never reused.

### dpmcore version

0.1.2rc13

### Python version

3.12.14

### Database backend

SQL Server 2019 (RTM-CU22) 15.0.4322.2

### OS

Ubuntu 24.04.4 LTS

### Additional Context

Ruled out during diagnosis:
- Python/pandas CPU — the process is idle while the query runs.
- SQL Server memory pressure — page life expectancy 3,984 s, data is in cache.

Database scale: TableVersionCell 429,414 rows · VariableVersion 295,912 ·
Cell 171,786 · TableVersionHeader 67,145 · HeaderVersion 51,650 · TableVersion 2,776.

Found while migrating a downstream consumer onto dpmcore.

andres-sole:
## Summary

Fixes #361

Validating an expression resolved every cell of a table through nine joined tables, once per cell selection. Two of those joins were the cost.

**The header of each axis was joined per cell.** A cell's row, column and sheet code came from an outer join whose `ON` clause was a disjunction over two different columns of two different tables — the `HeaderVersion` the table version pins, or, when it pins none, the one matching `HeaderID` directly. No optimiser can seek that, so it resolved the branch by scanning `HeaderVersion` once per driving row, three times over, once per axis. A table version has a few dozen headers against hundreds or thousands of cells, so the whole answer was being recomputed per cell.

**Module membership was joined per cell.** Whether a table version may be read is a property of that version, not of its cells, but the join made every cell appear once per module version carrying the table — fan-out that `DISTINCT` then collapsed again.

Both are now resolved away from the cells. The headers of a table version are read once (a few dozen rows) and applied to the cells in pandas; module membership is resolved once into the list of table versions to read. What is left is a scan of one table version's cells and the variable payload hanging off them.

**Impact**

Buffer pages touched per call, against a PostgreSQL copy of the 4.2.1 dictionary (394k `TableVersionCell` rows), summed over every statement the call issues:

| Table | Call | Before | After |
|---|---|---:|---:|
| C_18.00 | `get_axis_orders` | 2,673,757 | 30,845 |
| C_18.00 | `get_table_data` | 1,861,381 | 30,845 |
| C_18.00 | `get_filtered_datapoints` | 2,726,947 | 30,845 |
| C_13.01 | `get_axis_orders` | 1,185,908 | 31,698 |
| C_13.01 | `get_filtered_datapoints` | 1,237,581 | 31,667 |
| K_23.00 (median size) | `get_axis_orders` | 5,439 | 1,582 |
| K_23.00 (median size) | `get_filtered_datapoints` | 7,608 | 1,582 |
| C_106.02 (small) | `get_filtered_datapoints` | 1,519 | 678 |

## Checklist

## Impact / Risk


src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -2541,28 +2755,20 @@ def get_table_data(
...
+            if values is not None and values != ["*"]:
+                data = _narrow_to_dimension(
+                    data, code_col, values, axis_orders[axis]
+                )
 
+        data = _project(
+            _with_module_windows(data, membership), _TABLE_DATA_COLUMNS
+        )
         if len(data) > 0:
             data = data.sort_values("variable_id", na_position="last")
```
ruizmaa: `sort_values("variable_id")` defaults to `kind="quicksort"`, which isn't stable. When a cell is duplicated by two module versions with different release windows, ties on `variable_id` can swap order, so the following `drop_duplicates(keep="first")` may silently keep the wrong window. Suggest kind="stable" here to preserve the order `_with_module_windows` relies on.

src/dpmcore/dpm_xl/model_queries.py
```diff
@@ -2103,46 +2103,253 @@ class ViewDatapointsQuery:
...
+    ] = {}
+
+    # The module versions a (table, release, live) scope resolves to,
+    # and with them the table versions to read. Resolving one costs up
+    # to four queries, and every cell selection over the same table
+    # asked for the same answer.
+    _MEMBERSHIP_CACHE: dict[
+        tuple[Hashable, str, int | None, bool, bool],
+        pd.DataFrame,
+    ] = {}
```
ruizmaa, line 2122: This PR adds `_CELL_FRAME_CACHE` and `_MEMBERSHIP_CACHE`, but the `_clear_query_caches` fixtures in `test_live_table_versions.py` and `test_calculations_export.py` only clear `_TABLE_DATA_CACHE` and `_AXIS_ORDER_CACHE`. I think it would be nice to add `.clear()` for the two new caches in both fixtures too.

andres-sole: fixed. Both test fixtures now clear all caches

src/dpmcore/dpm_xl/model_queries.py
andres-sole: Fixed with `kind="stable"`, with a comment explaining what the sort is for.

ruizmaa (approved):

